PR #11, opened 2015-03-19
Spell check & name check

luminairre:
lines tagged with //namecheck seems to call for same name, but written
in different name. since we don't have access to the game or database,
please help to re-check those lines.

thanks!


imcgames:
Thank you! :)

PR #13, opened 2015-03-19
Update ETC_en.tsv

Saklas:
added line 103 translation


imcgames:
Not merged but manually added :)

PR #14, opened 2015-03-19
Job requirement

luminairre:
Fixed job requirement and names according to english text from the site,
it seems "yen" and "circle" refers to same thing, and we currently don't
know what's the term of the requirement, is it "grade"? or "level"? but
these texts seems to be pointing to a requirement for skill or advancing
to next class

ETC_20150312_001869 says "Mugosa" but I don't know what class is "무고사"
exactly.


Saklas:
according to the wiki, Mugosa is a monster:
http://www.tosbase.com/database/monsters/57233/

imcgames:
The wiki is wrong. that is "무고사 마스터" not "무고사". 
"무고사" is one of class in TOS.

luminairre:
found it, it's "Wugushi"

PR #19, opened 2015-03-19
Update QUEST_LV_0100_en.tsv

Waterfighter:
I cannot read or understand Korean, so any changes I made are based upon my understanding of what was written in English.


imcgames:
Thank you :)

PR #27, opened 2015-03-20
Update ETC_en.tsv

healthytofu:


Sunaries:
???

You added what has already been there. Korean shouldn't be present in the file where it should be English.

vegax87:
healthytofu, you need to update your files first, these files are very old now...

ensata:
It's not just the files. Everything is old. This is a pull request from March. Was it reopened on accident?

imcgames:
We will close this pull request as the files seem to be from the old version. 
You may resend us the edits using the latest files.  :wink:

PR #32, opened 2015-03-20
Update ETC_en.tsv

Saklas:
added lines 236-237 translations and corrected line 238 to fit name:

Austeza was probably just typo, z-y switch up on the keyboard :)?


imcgames:
Actually, it is Austeja, one of the Lithuaninan Goddesses :)

PR #42, opened 2015-03-21
Update ETC_en.tsv

ShinHayato:


vegax87:
I agree, I checked lines some lines and they're unreadable because they're copy-pasted from Google without a minimal adaptation.
When you see something like {img YOUR_BUTTON 40 40} you need to translate like this:
"Use the {img YOUR_BUTTON 40 40} key" and not "Use the key {img YOUR_BUTTON 40 40}"
I manually fixed line 811 in my pull request

imcgames:
We closed this pull request refer to guys.

PR #43, opened 2015-03-21
Update ETC_en.tsv

AEJung:
Most of the original Korean is incomplete. So, the English is fragmented and missing key words. I assume the coding slots in interchangeable words in those places?


imcgames:
Thank you :)

PR #49, opened 2015-03-21
Update QUEST_LV_0100_en.tsv

Nagoda:
Revising already translated text for gammar and more natural flow.

A person's official title should always be capitalized, such as "Sir Kayetonas" or "Brother Baidutis".

Judging from the context, "Devils" seems like it should be a collective proper noun ("The Devils" as a recognizable enemy faction), and probably should be capitalized to refer to them as a whole group. I didn't capitalize every instance of Devils that I saw, but I did change the ones near the lines I was already editing. Some of them were already capitalized and didn't need changing.

 I have some doubts whether "Devil" is the best word to use here at all, but I can't read the original Korean and it would require some discussion to go and change the name of an entire group of enemies.


bavarois:
I think "Demon" is a better fit in this case.  The word "Devil" usually refers to "악마" rather than "마족".

PR #59, opened 2015-03-23
Update QUEST_LV_0100_en.tsv

Nagoda:
After some discussion, did a find>replace on every instance of the word "Devil" and changed it to "Demon".

Minor editing. Some minor and a couple moderate rewrites for more natural flow.

At this point I've finished a first pass skimming through the whole Quest file. There's still some stuff I could clean up, but there's not a ton left that I'm comfortable editing without more story/game context, or being able to read the original script.

There's at least a few dozen lines of gibberish from machine translators in the file that are totally beyond salvaging. I may try to compile a list and start an Issue discussion about what should be done with them (probably just deleted).


imcgames:
:+1:

PR #62, opened 2015-03-23
Update ETC_en.tsv

vegax87:
Updated lines 835,836,838,840,842-844,860,871,942,958

Added lines: 809,811,813,815,818,819,821,827,828,829,832,883-888,891,892


imcgames:
:+1:

PR #69, opened 2015-03-24
Update ETC_en.tsv

MarkofKain:
Line 758 Replaced Avoidability with Evasion. Or is it called Dodge?


Sunaries:
Avoidability, evasion and dodge are similar words and mean the same basically.

imcgames:
:+1:

PR #70, opened 2015-03-24
Update ETC_en.tsv

vegax87:
Fixed some translations and added some lines

Note: I'm not sure about ETC_20150312_002352, Bing translates it "Buy Purple Wings" because purple in korean is 보라색 (bo-ra-saek); 자주색 (ja-ju-saek); 자주 [紫朱](ja-ju) - purple


Sunaries:
Bing Translate should never be trusted as it is done by a machine. We don't know if it is an NPC or object. Therefore your edit in ETC_20150312_002352 should not be merged.

imcgames:
This is not enough for merge. Sorry.!!

PR #78, opened 2015-03-25
Update ETC_en.tsv

mickeyjulian:
ETC_20150312_002191 Modify Visibility

ETC_20150312_002198 Slowed movement and mental strength is reduced.

ETC_20150312_002200 A state of continuous equipment damage.

ETC_20150312_002203 Defense power was reduced but attack power increased.

ETC_20150312_002238 Sticky Foam


imcgames:
:+1:

PR #80, opened 2015-03-25
Update ETC_en.tsv

mickeyjulian:
ETC_20150312_002262 A state of continuous flame damage.

ETC_20150312_002376 HP / SP / stamina consumed can be recovered by using the following functions.

ETC_20150312_002375 Particular class specialization can check additional information of monsters.

ETC_20150312_002374 In monster information window , you can check the level and the name of the chosen monster and how strong the monster is.

ETC_20150312_002373 You can look at the monster , you will be able to view the monster information window as it approaches.


imcgames:
:+1:

mickeyjulian:
Thanks :)

PR #81, opened 2015-03-25
Update QUEST_LV_0100_en.tsv

ShinHayato:
add line 6266, 6288, 6295, 6299, 6301, 6309, 6314-6315, 6379


imcgames:
This update is not accurate compared to original text.
Thank you.

PR #112, opened 2015-03-26
Update ITEM_en.tsv

Sunaries:
Updated & corrected few names of items and weapons. Lots of work needs to be done. 

I am aware of the prefix 조악한, 로우 etc. If you know a better English term with the same meaning, try it out.


vegax87:
you could use Rugged
for example: Rugged Leather Armor

PR #131, opened 2015-03-30
Update ETC_en.tsv

Sunaries:
What happened here??? I see many blank lines that were translated before.... It feels demotivating to see this many lines gone..


imcgames:
According to your feedback like issue, we updated new clean files on Github.
Thank you :+1:

PR #132, opened 2015-03-30
Update ETC_en.tsv

MasterCyria:
Updated lines 1-22


MasterCyria:
Reminder to myself: On line 1 it should say an incense, not a.

vegax87:
you can make a second commit here to fix line 1

MasterCyria:
@vegax87 Thanks, I'm a bit new to GitHub :)

MasterCyria:
New update: Added up to line 35

MasterCyria:
A few notes:

Line 28's translation seems off from what I got, but my translation didn't make any sense in English so I didn't change it

Line 33: The term Elyos needs to be checked

Line 34: Uncertain what this references. I assume a zone based on line 35. It may be better as "The Wizard's Tower", but I would need more context.

Line 23: Term check on Bearkars. If it's a name and not a plural name of a mob, line should be changed to "Bearkar's suppressed power returned."

Sunaries:
Line 23: Bearkars → Bearkaras. It is not plural since it is a boss monster. So, it should be: The suppressed power of Bearkaras has returned.
Line 28: 신수 has a variety of meaning. It can mean Divine Beast, Shinsoo, God etc. The line would be somewhere along the lines of: Ask about the day of Shinsoo of four years ago. 
Line 33: Elyos → Demons
Line 34: 'The Magician's Tower' is indeed an area/zone. 
Line 35: Village of the Damned → Cursed City

MasterCyria:
Updated with the clarifications from @Sunaries, except for Line 35. I think that one may be a personal preference, similar to prophet debate. I think they both make sense, depending on the size and style of the zone. I imagine a city to be more industrial, while a village is more country side and medieval feeling.

Probably would need to take a look at the zone in ToS to determine.

Sunaries:
You forgot the include 'of' after 'Shinsoo' at line 28. 
Reading the text, 'The Cursed City' makes more sense for 저주받은 도시. 
도시 is city, while 마을 means village. 

It ultimately comes down to @imcgames since they know the game better than us. I have no idea what the map looks like.

MasterCyria:
I'll change it to the Cursed City also - thanks for explaining! :)

For line 28, the Day of Shinsoo of four years ago doesn't actually read correctly, English wise - but I know the intent; the current wording implies that you're going back in time and asking it four years ago :) I'm thinking maybe "Ask about the Day of Shinsoo that occurred four years ago."

Sunaries:
:+1: ^^

imcgames:
Thank you! :+1:

PR #145, opened 2015-04-01
Update ETC_en.tsv

ruru1:
I don't know the unique name like "000 town", just make it myself :)


vegax87:
you can check here the names of the locations: https://github.com/Treeofsavior/EnglishTranslation/wiki/2.2.%20Map
if you find a familiar name, copy it from the Wiki

PR #147, opened 2015-04-01
Update ITEM_en.tsv

vegax87:
Corrected 2 item names


imcgames:
I think Status is more accurate compared to original text "스탯".
"스탯" is similar to "스테이터스"

The other one updated via manual step.
Thank you :+1:

PR #189, opened 2015-04-09
Update QUEST_LV_0100_en.tsv

xapher19:
I made some more grammatical corrections, to better the flow of the sentences.

### DIFF
--- a/QUEST_LV_0100_en.tsv
+++ b/QUEST_LV_0100_en.tsv
@@ -40,11 +40,11 @@ QUEST_LV_0100_20150317_000039	You must be busy for other things so thank you so 
 QUEST_LV_0100_20150317_000040	Secret Guardian
 QUEST_LV_0100_20150317_000041	The goddess already forecasted this.{nl}All beings can not escape from forecasting.
 QUEST_LV_0100_20150317_000042	Liaison Officer Niels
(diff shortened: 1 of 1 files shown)


vegax87:
In QUEST_LV_0100_20150317_000013     you deleted a {nl} tag after "I warn you!" , it's a tag written by devs to define a new line, you shouldn't remove these tags

xapher19:
I changed it as per the [wiki](https://github.com/Treeofsavior/EnglishTranslation/wiki). There it notes that the new line tags can be moved around as new lines in Korean may not be needed in English. If the devs wish it to remain, we can replace it.

Sunaries:
'Black' in 'Soldier Black' is wrong. Vlock or Bloc is a better name for this soldier.

xapher19:
But 블락 doesn't sound anything like Vlock or Bloc. ㅏ doesn't make an "o" sound it's pronounced like the "a" in "apple", also ㅂ is pronounced as a merge between a b and a p sound as in "ball" or "Paul". That's why I translated it as Black.

Sunaries:
Black would be 블랙. I'd go with Blak then.

xapher19:
That's a fair point and a good compromise. It's been edited.

imcgames:
:+1:

PR #209, opened 2015-04-13
Update ITEM_en.tsv

Sunaries:
- To avoid confusion on the term 'Attribute' and 'Status', I suggest to replace 'Status Point' with 'Ability Point'


imcgames:
:+1:

PR #210, opened 2015-04-13
Update ETC_en.tsv

Sunaries:
- To avoid confusion on the term 'Attribute' and 'Status', I suggest to replace 'Status Point' with 'Ability Point' 
- Updated NPC names according to the Wiki


imcgames:
:+1:

PR #224, opened 2015-04-16
Fixed inconsistencies with proper nouns on wiki.

dlam:
Changes of note:
- Even though "Nukella" only has one L in the wiki; the item's icon from the game's wiki has 2 L's and it's clearly a play on the famous brand "Nutella", so I kept it with 2 L's instead of 1. Icon found here: http://toswiki.treeofsaviorgame.com/_media/item/misc/icon_item_nukela.png?h=68&tok=805d8c
- Changed "Biscuits" to "Hardtack" this is an awkward change, but it turns out [Hardtack is a thing](http://en.wikipedia.org/wiki/Hardtack).

After doing a couple of these, I noticed some of the translations on the wiki are somewhat awkward (like using Raudonelis vs Oregano), so I'm going to stop prematurely and make an issue about proper item names before doing a complete sweep.


imcgames:
Thank you for first contribution for TOS.
We're looking forward to seeing you again. :+1:

PR #225, opened 2015-04-16
Update ETC_en.tsv

Sunaries:
Minor edit. It is annoying to edit through Github, because there is this some sort of lag for me whenever I edit.


imcgames:
Thank you always. :+1:

PR #233, opened 2015-04-17
Fix spacing in ITEM_en.tsv

dlam:
Some lines were changed by fan translators from tab spacing to 2 spaces (most likely due to github editor).

This change fixes all spaces between item codes and translations to tab spacing, for consistency with the original source.

This was bothering me...


imcgames:
:+1:

PR #238, opened 2015-04-17
Update ITEM_en.tsv

Sunaries:
Minor update.

### DIFF
--- a/ITEM_en.tsv
+++ b/ITEM_en.tsv
@@ -2382,25 +2382,25 @@ ITEM_20150317_002381	Zignuts Fruit
 ITEM_20150317_002382	Zignuts Seed
 ITEM_20150317_002383	Panto Leaf Decoration
 ITEM_20150317_002384	Wing of Humming Bird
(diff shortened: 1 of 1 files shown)


ITEM_en.tsv
```diff
@@ -2458,7 +2458,7 @@ ITEM_20150317_002457	Pokuborn Meat
 ITEM_20150317_002458	
 ITEM_20150317_002459	Loxodon Meat
 ITEM_20150317_002460	
-ITEM_20150317_002461	Ansr
+ITEM_20150317_002461	Ansur
```
dlam, line 2461: 안스르 -> Ansr

From the [wiki](https://github.com/Treeofsavior/EnglishTranslation/wiki/2.4.%20Item)

Sunaries:
The Wiki isn't always right since it was updated a few times. Ansur sounds better than Ansr.

dlam:
I agree that the wiki isn't always right, but it's important to be consistent... should probably open an issue to update the wiki with the preferred romanization of 안스르.

imcgames:
Updated "Anusr" in Wiki

PR #240, opened 2015-04-18
Proofread ITEM_20150317_000000 - ITEM_20150317_000100.

dlam:
Decided to fix up some grammatical mistakes and inconsistencies with official translations from the wiki.


Sunaries:
Aras Falchion -> Aras's/Aras' Falchion as it indicates it is a falchion from Aras. 
e.g. Mayor's Falchion

dlam:
Is Aras a person or a place / species?

In the latter context it's grammatically correct to omit the apostrophe, because "Aras" is being used as an adjective. Although not explicitly necessary, if Aras is a place it should probably be changed to something like "Arasian".

Anyways, I'm just going off the official translations from the [wiki](https://github.com/Treeofsavior/EnglishTranslation/wiki/2.3.%20Equipment), but can open an issue to fix if Aras is a person.

Sunaries:
The question about what Aras is should be asked to @imcgames. I don't know much about the story yet. 

The Wiki has quite some inconsistencies. Such as: Godendag (Goedendag), Morning Star (Morningstar) etc.

dlam:
@Sunaries Yep, I completely agree with you. Time to wait for @imcgames to reply.

imcgames:
Here is an answer to your question. :)

dlam:
It should probably be Aras' [Weapon] in this case.

PR #242, opened 2015-04-18
Fixed all {nl} tags

vegax87:
Fixed all {nl} tags which weren't lowercase


Kaosu-san:
On line 1620 there is a "mosnter" that need to be corrected, otherwise it's a pretty good job!
:+1:

vegax87:
Fixed line 1620

imcgames:
:+1:

PR #249, opened 2015-04-20
Update QUEST_JOBSTEP_en.tsv

vegax87:
fixed Pardoner Advancement


imcgames:
Updated through a manual step :+1:

PR #250, opened 2015-04-20
Update ETC_en.tsv

Sunaries:
It is becoming tedious to edit in large files like this one. It slows down typing. Takes me approximately 10 seconds to edit a line. :/


imcgames:
:+1:

PR #274, opened 2015-04-25
Update ITEM_en.tsv

xMiiNTY:
I'm assuming what I've done are monster drops. All names taken from etc file.


imcgames:
Thank you :+1:

PR #280, opened 2015-04-27
Update ETC_en.tsv

TheStargrazer:
ETC_en.tsv lines 1-502
1. Fixed some of the grammar and capitalization errors and reread the names.
2. Rephrased some dialogue to make it sound more natural while staying true to the original idea.
3. I decided to leave the more vague lines as they are so someone can provide a better translation.


imcgames:
Thank you so much :+1:

PR #282, opened 2015-04-28
Update ITEM_en.tsv

xMiiNTY:
I was translating this halfway until I found out that it was updated it. So I've merged the two together myself and fixed most of his errors such as some capitalizing of words, spelling of words (Obtained rather than Obtaiend), some correction of Hangeul vs Romanization (Cysts rather than Sisti) and corrected some order and spelling of words to maintain consistency (King Tomb rather than Royal Tomb and also some monster names). I've also fixed alot of errors that translator machines would typically make.


imcgames:
Thank you :+1:

PR #309, opened 2015-05-14
Update QUEST_LV_0200_en.tsv

zainezed01:
Revised and added more translations.

Are the npc names finalized in this translation? For example, in the case of "Commander O'Rourke" did imcgames officially translate this name or someone else? I'd like some insight as to who translated the names so I know which translation to use later on.  Is it possible to create a reference to wiki for npc names since I'm talking about this?


imcgames:
The NPC names are not yet finalized. We updated it using the machine translation for a quick update.
If you guys have any suggestions for names, then please update through Issues or Pull request.
It will be updated after checking whether it's good or not compared to TOS's worldview.
And we do plan to make the edits public in the future when the game is released, so please hang on until then :+1: 

Thank you.

PR #318, opened 2015-05-15
Update QUEST_LV_0200_en.tsv

zainezed01:
added more translations.

don't know if this was intentional but line 63 appears to be missing from both the raw and eng translation page.


imcgames:
The 63 line is no longer used in game, so deleted.

Thank you :+1:

PR #344, opened 2015-05-19
Update ITEM_en.tsv

emailboxu:
Armor string updates. Mostly naturalization of English.

Also some questions regarding translating
1. ln 922, it says "The comfortable Veris Boots" but the string before it is "Scale Mail". Obviously not a programmer but I thought I'd point out that this string doesn't directly correlate with the line above it. 

Personally I think knowing which strings match which item would be beneficial as many "this" or "its" could be changed to "these boots" or "these gloves", etc. It would make the description strings more natural. For example, 928 (if referring to gloves), instead of "Swordsmen like to wear this into real battles", changed to "Swordsmen like to wear these gloves into real battles." 

Some lines I added equipment specifics to (based on ToSBase's Equipment database) - 1064, 1066, 1068, 1070
1. Are parenthesis (or brackets) allowed to be used in description strings or will that mess with the programming? In line 940, "like these pants" is ideally put in parenthesis rather than commas (better flow/more natural).
2. Some strings are used for multiple items (line 980, based on this site: http://www.tosbase.com/database/items/equipment/), but this doesn't work well because Gloves, Boots, and Pants are plural ("THESE pants ARE light and comfortable) while Armor is singular ("THIS hauberk IS light and comfortable"). If the devs could put in separate strings for Gloves/Boots/Pants and Armor (assuming top armor) it would be useful and more natural in-game. Currently it seems like the previous translation just made everything singular but it doesn't really work like that in English.

### DIFF
--- a/ITEM_en.tsv
+++ b/ITEM_en.tsv
@@ -1205,20 +1205,20 @@ ITEM_20150317_001298	Defense Necklace
 ITEM_20150317_001299	Novice Necklace
 ITEM_20150317_001300	The fact that such equipment is so affordable is a blessing for beginners.
 ITEM_20150317_001301	Vubbe Talisman
-ITEM_20150317_001302	This is the reason why attacks sometimes have no effect on Vubbes.
+ITEM_20150317_001302	This talisman is the reason why attacks sometimes have no effect on Vubbes.
 ITEM_20150317_001303	Chupacabra Necklace
-ITEM_20150317_001304	Rumor says that the teeth of Chupacabra hiss to each other while chattering. Regardless, they definitely contain some magic power.
-ITEM_20150317_001306	It's not recommended to take it off while you're carrying something.
+ITEM_20150317_001304	Rumor says that the teeth of Chupacabra whisper to each other while chattering. Regardless, they definitely contain some magic power.
+ITEM_20150317_001306	It's not recommended to take this off while you're carrying something.
 ITEM_20150317_001307	Iron Necklace
-ITEM_20150317_001308	You'd feel you became lighter, but in reality you did not become any thinner.
+ITEM_20150317_001308	You might feel like you became lighter, but in reality you will not have become any thinner.
 ITEM_20150317_001309	Steel Necklace
-ITEM_20150317_001310	In battle, a small factor may make the difference between life and death. You certainly won't regret by wearing this necklace.
+ITEM_20150317_001310	In battle, a small factor may make the difference between life and death. You certainly won't regret wearing this necklace.
 ITEM_20150317_001311	Talisman
 ITEM_20150317_001312	Sadly it won't help you much when studying.
 ITEM_20150317_001313	Klaipeda Chain
 ITEM_20150317_001314	This necklace brings out the power of the wearer's soul.
 ITEM_20150317_001315	Might Fedimian Chain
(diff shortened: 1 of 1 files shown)


ITEM_en.tsv
```diff
@@ -859,125 +859,125 @@ ITEM_20150317_000930	Find something else to wear in the highlands. It's cold.
...
+ITEM_20150317_000980	If you don't have an aversion to insects, this is a good piece of armor to not only protect yourself, but also to hold a weapon.
 ITEM_20150317_000981	Veris Boots
-ITEM_20150317_000982	A swordsman's boots must be as sturdy as this one. Not only because of the enemies, but also to guard against your own weapon dropping on your foot.
+ITEM_20150317_000982	A swordsman's boots must be at least as sturdy as this one. This is not only to protect him from his enemies, but also to guard against him dropping his own weapon on his feet.
 ITEM_20150317_000983	Veris Pants
-ITEM_20150317_000984	Some warriors believe that the materials used for this item will attract its vengeful kins.
+ITEM_20150317_000984	Some warriors believe that the materials used for this item will attract its vengeful kin.
 ITEM_20150317_000985	Veris Tunic
-ITEM_20150317_000986	Equipment that offers protection superior to the plate armor by strengthening the shells of an insectoid monster.
+ITEM_20150317_000986	This equipment that offer superior protection to plate armor by strengthening it with the shells of an insectoid monster.
```
Sunaries: Change it to: _This equipment offers superior..._ 
Sounds more natural.

ITEM_en.tsv
```diff
@@ -840,18 +840,18 @@ ITEM_20150317_000911	Raganos Horn
...
-ITEM_20150317_000918	A shoes that have been worn in some areas and terrain of the kingdom. If not wear anything, it is trouble.
+ITEM_20150317_000918	A pair of shoes that can worn in a variety of areas and terrain throughout the kingdom. Not wearing shoes is quite dangerous.
 ITEM_20150317_000919	Light Pants
 ITEM_20150317_000920	Pants
-ITEM_20150317_000921	There is no adventure up to path without it.
+ITEM_20150317_000921	There are no adventurers who journey out without these.
 ITEM_20150317_000922	Light Armor
 ITEM_20150317_000923	Top [Leather]
-ITEM_20150317_000924	As long as it protects you, it doesn't matter what kind of leather it is made of.
+ITEM_20150317_000924	Tt doesn't matter what kind of leather it is made of as long as it protects the wearer.
```
Sunaries: Tt -> It

ITEM_en.tsv
emailboxu: Thanks, missed the "that" in the middle there.

ITEM_en.tsv
emailboxu: P.S., are you the real Spadow of Maplestory fame? Huge fan, haha.

imcgames:
Thank you for your suggestions. We do try to have the translations lines correlated as much as possible but sometimes there are factors that affect it like game data, translations build, order, etc for which we have to make some compromises which could lead to a break in the correlation of the translation strings. 

As for translation strings, we do not recommend the use of symbols and characters other than alphabets and numbers unless it's also found in the original text. You may use parenthesis only if it is really necessary but as for brackets, we highly discourage you from using it as it may affect the game. :smile:

ITEM_en.tsv
Sunaries: Yes, I am. There can only be one Spadow. ;-)

Issue #354, opened 2015-05-20
Plurals?


emailboxu:
Going through the various monster drops and am wondering if I need to leave monsters as singular. 

> Looks similar to a bat wing, but is actually a plant stem. Obtained from Dandels.

Makes more grammatical sense since Dandels are (presumably) a common monster and you can farm the said Item (Dandel Wing).

The alternative is to leave them singular:

> Looks similar to a bat wing, but is actually a plant stem. Obtained from Dandel.

which sounds weird. Let me know what you think, I'm currently just pluralizing everything but it won't be hard to undo that in another patch.

megaoka:
Either

> Obtained from Dandel**s**

or

> Obtained from **a** Dandel

will work in this situation. Personally though, I prefer the first. "Obtained from Dandel" should only be used when it's referring to a specific person or object, like a named NPC or Boss.

megaoka:
On a similar note, I have been changing things from "X of Y" to "Y's X" because I think it sounds better and is shorter/neater. Is there any case where we should not have it like this?

e.g. - "Bone of Popolion" -> "Popolion's Bone"

emailboxu:
I switched all of the X of Ys to Y X (Ie, Popolion Bone). Y's X usually refers to a proper noun for Y (ie, emailboxu's post) rather than a generic Y X (ie, dog treat). I figured since all of these strings are generic mobs/not bosses Y X would be better than Y's X. Again this is fairly easy to change. 

I agree that the plurals sound much better for non-NPC/Boss related items, though.

megaoka:
+1

emailboxu:
To add to the X of Y thing, all of the possible permutations (X of Y, Y X, Y's X) are written exactly the same way in Korean, which is why we're doing it for anyone curious. X of Y is the closest "literal" translation but often sounds odd for natural English speakers.

megaoka:
Right. Literal translations are nice, but often do not work out. It's best if you get the concept and adapt it to the language. This is my opinion, of course.

I agree with the Y X instead of Y's X, that makes sense now that you explain it. :+1:

emailboxu:
:+1:

PR #358, opened 2015-05-21
Update ITEM_en.tsv

megaoka:
Minor changes.

Made a few fixes as per the discussion here: https://github.com/Treeofsavior/EnglishTranslation/issues/354

Unsure about lines 2285 and 2286. Is Violet a monster, named/boss monster, or NPC? "Violet Flower" sounds too much like "a flower that is violet"

(and corrected a typo I made)

### DIFF
--- a/ITEM_en.tsv
+++ b/ITEM_en.tsv
@@ -2242,7 +2242,7 @@ ITEM_20150317_002358	Jukopus Rope
 ITEM_20150317_002359	Red Kepa Stem
 ITEM_20150317_002360	Vubbe Spear Head
 ITEM_20150317_002361	Vubbe Blood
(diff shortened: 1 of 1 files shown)


emailboxu:
Actually 2286 should be Violet's Crystal, I should have checked with the KR file when I translated that. There's no 'of' for Violet Flower but I'm guessing it's the same.

emailboxu:
Unless it's a common mob. Either way might be safer to leave the possessive in there.

megaoka:
I wasn't  sure, because I didn't see "Violet" in the mobs list, but it's also strange because it's already a noun.. I think the possessive is a safer change if you want to make it.

imcgames:
Violet is just a monster :+1:

PR #360, opened 2015-05-22
Minor Fixes

TOSBase:
Added some missing tabs, changed ' - ...' parameters like they are in the korean file with 2 spaces.

Note: the ' - ...' lines are not normal item descriptions but more like item attributes.

http://www.tosbase.com/database/items/223104/
vs.
http://www.tosbase.com/database/items/162114/


imcgames:
Thank you for your suggestion. :+1:

PR #367, opened 2015-05-28
Update QUEST_LV_0200_en.tsv

megaoka:
More name changes, unsure about the Pyromancer Master's gender. Let us play CBT so we can translate better! :P


imcgames:
:+1:

PR #370, opened 2015-05-29
Update QUEST_JOBSTEP_en.tsv

Trepoleth:
Line 171 it's either "of" or "to".
Line 173 it's either "gratitude" or "gratuity".


megaoka:
Both lines 171 and 173 look correct. Nice work. :+1:

imcgames:
Thank you :smile:

PR #387, opened 2015-06-04
Update ITEM_en.tsv

emailboxu:
3529 - Fire -> Aura as recommended

3587 - 칼리스 기단 ? What does 기단 refer to here?

3591 - What does 사념체 mean in this context? I understood it to be spirits.

~3600.


imcgames:
Thank you very much :+1: 

3587 - 칼리스 기단 ? What does 기단 refer to here?
=> 기단 seems like a typo of 기사단

3591 - What does 사념체 mean in this context? I understood it to be spirits.
=>Yup, that's right. It's something like spirits with sinister or malicious intents. :smiley:

PR #391, opened 2015-06-08
Update ITEM_en.tsv

emailboxu:
3629 - Not sure what silk sign/token is referring to, probably another item?

~3650


imcgames:
Yes, It's another item :+1:

PR #397, opened 2015-06-11
Update ITEM_en.tsv

emailboxu:
3778, 3779, don't know what 마군주 means. 마군 appears to be referring to demons (?) and 주, master? Alternatively, 마, Demonic/Magic, 군주, prince, monarch, etc. Also in line 3784.

3795~3798, I'm pretty sure 포마인 refers to an in-game monster, item, NPC, place, etc. of some kind, not "Four main," but it's not in the Wiki. Anyone know what it is?

~3800.


imcgames:
The meaning of "마" is "Demons" and "군주" is "Master".
And "포마인" is item name.

:+1:

PR #404, opened 2015-06-16
Update ITEM_en.tsv

emailboxu:
3862 - Think there's something missing from here? The original Korean doesn't make much sense.

~3900. Will probably try to get through 3901~4100 tomorrow, since it's mostly item names and won't take long.


imcgames:
Thank you :)

PR #427, opened 2015-06-30
Update ITEM_en.tsv

Sourpusss:
Edited Necklaces, Bracelets, and Costumes (I only changed one small thing in the costume section, but I proofed it all).

Line 1238 was kind of a mouthful, so I think I worked into something a bit more cohesive without taking out anything.

Lines 1342 through 1344 seem like they're out of order? I don't want to touch them in case they're matching up with different items, but otherwise it seems like lines 1342 and 1344 should be combined, and then 1343 follows.


imcgames:
Thank you :smile:

Issue #120, opened 2015-03-27
Revelator vs Oracle/Prophet


aentan:
계시자로 is more like oracle or prophet. Revelator just sounds weird. Would be good to know the context.

Nagoda:
This issue was brought up in the Word/Phrase translation thread here, https://github.com/Treeofsavior/EnglishTranslation/issues/50

Ultimately we didn't come to a strong conclusion and felt we needed more context. I personally think Prophet sounds the best, whether or not it's most accurate. If you have additional thoughts to add, feel free.

Issue #50, opened 2015-03-22
Individual word/phrase translations


Nagoda:
I don't speak Korean so my ability to help is limited, but I notice when something sounds goofy in English. I've come across a couple words that are used a lot in the Quest file as translated so far which seem like they might not be quite right.

The first word that appears a lot is "Devil", which seems to be a major enemy type that appears many many times in the Quest dialog, and players will no doubt be fighting them a bunch. I believe the original Korean word used is "마족". The most common usage of "Devil" in English is to refer directly to Satan ("the" Devil). When it's used to refer to lesser evil spirits, it often has a kind of mischievous or impish sort of connotation. It's simply not used very often in games media to describe evil or unholy monsters.

I mentioned this in one of my pull requests and another user suggested "Demon" as a better choice. I lack the context needed to know if Demon is a more accurate translation, but it certainly seems to sound better.

The second odd word is "Revelator". The Quest file makes many mentions of "Revelations", "Revelators", and seems to refer to the player character directly as being "the Revelator", as some kind of "chosen one" title. The original Korean word seems to be "계시" or "계시자". My main problem with Revelator is that it's a very seldom used and obscure word, to the point where it ends up just sounding like a too literal translation.

I wonder if "Prophet" might be a more useful title that has more or less the same meaning as Revelator. It would be helpful if someone could provide better context for what the original script is trying to say.

AEJung:
That's riiight, it is referring to the player character. There was some quest dialog from the first G-Star (2013) video that talks about how your character had been visited by "the Goddess" (Laima) in their dreams and told to journey to Klaipeda (probably among other things). As for an actual title for it...Revelator...Prophet... When you really delve into the origins and meanings of the words they both work well, but I think when most people thinking of a prophet is someone who is predicting the future based on what God told them (or, actually, gods -- I think more of Greek mythology when I hear "prophet" instead of Christian)? Whereas a revelation tended to be more of a task or mission that God wanted them to complete. So Revelator is probably just a tiny bit more appropriate based on popular knowledge. 

Ah and yeah, demon is probably more appropriate. 마족 is Chinese-derived Korean, 마 (ma) meaning evil, 족 (jok) meaning tribe. I mention the Chinese part because it can be hard to find correct definitions for them in Korean dictionaries. 

Hope that helped!

MarkofKain:
[...]

PR #441, opened 2015-07-02
Update ETC.tsv

TendrilTender:
Fixed some sentences and typos. Changed all instances of Revelator to Oracle.

### DIFF
--- a/ETC.tsv
+++ b/ETC.tsv
@@ -10518,12 +10518,12 @@ ETC_20150428_011497	{character3} of team obtained {count3}silver
 ETC_20150428_011498	You can not give up in progress or complete quest standby.
 ETC_20150428_011499	Warp Movement Area Ranking
 ETC_20150428_011500	Monster Kill Area Ranking
-ETC_20150428_011501	Oracle Combat Disabled Area Ranking
-ETC_20150428_011502	Oracle Combat Disabled Area Ranking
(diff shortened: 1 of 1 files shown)


Chaosbelow:
Revelator is a term for the player character not the name of the class in this instance. This was reffered to in issue #50 and #120 but we need more context. For now 계시자 translates directly to Revelator so leave it at that.

TendrilTender:
Noted eith the Oracle thing, changed it back in the places it doesn't refer to the class. I had failed to realize the context on the Line about harugals, changed it to your version. Where is the spacing wrong though? Can't seem to find it.

Thanks for the feedback!

Chaosbelow:
Great work, keep it up. Line 11309: 
+ETC_20150428_011309_The mysterious instead of 
+ETC_20150428_011309___ The mysterious

The underscore is represented as space. 
Notice the indentation? Match it up with the rest of the lines before and after it. 
Sorry if its a bad explanation.

TendrilTender:
Hmm, I can see that it's wrong when I look at the changed lines, but when i look at the actual file it looks aligned correctly. Not sure what to do about this.

TendrilTender:
There, fixed it. Not sure what the cause was, so I just copied over the old line and used that instead.

Chaosbelow:
:+1: Great work!

imcgames:
Thank you :smile:

Issue #299, opened 2015-05-03
Differences between the official reference wiki entries and the merged translations


hasenbanck:
I have progressed with my translation tool now so much, that I can search for standard terms in the TSV files and locate inconsistent usage of those. Right now, I get my reference data from the official wiki entries. But those entries stand in contrast with the merged translations. I think we need to find a way to fix those entries in the reference wiki ASAP, or find anther way, to create a stable reference data pool (Items, Monster, NPC names etc.).

http://pastebin.com/409fVfJk

Explanation:
Source text = Korean original
Target text = English translation
Standard term = Name as found in the reference wiki
TSV-entry = Name as found in the English translations

Some log entries don't have a TSV-entry. In that case the word was used in a longer context and you have to search for it inside the original TSV files.

How can we resolve these inconsistencies?

Sunaries:
The official wiki contains inconsistencies. For example: 
- 무고사 (Mugosa) = Wugushi. But in the wiki it is referred as 우구시. 
- 독수리깃 is Eagle Feather and not Eagle Git.
- "Two-Hand Sword" should be "Two-Handed Sword", etc.
-  "Recipe-Cane" should be "Recipe - Cane", etc.

The _TSV-entries_ contains higher quality translations , but some of the _standard terms_ are good. 
The wiki should be updated to resolve some of your issues. However, the wiki only provides a limited amount of info compared to the files. Most of the correctly translated text and descriptions are in the _TSV-entries_. 

Only @imcgames can edit the wiki, but nobody else can. I think it would be wise if they allow members to edit the wiki, so that the terms and names can be updated to a better quality and the appropriate translation. All they need to do is give feedback on it and check if its alright or not.

hasenbanck:
My goal is to unify all standard terms in the TSV files. Right now we have no way to track those standard terms, since the ones in the wiki is are wrong on many occasions and the entries in the TSV files are not finished yet. I could fix the entries from the wiki myself in my own code repository (using the TSV -entries), but I think it would be better to fix the wiki entries. 

It's also near to impossible to know "which" entries are finished. QA is kinda impossible right now.

imcgames:
UI for items can sometimes be limited and that is why we try to keep some item names as short as we can. 

As for the wiki edits, it is limited to prevent possible trolling activities so for the time being, we will apply the changes in wiki when suggestions are made. But we do plan to make the edits public in the future when the game is released so please hang on until then :smile:

[...]

PR #442, opened 2015-07-02
Update QUEST_JOBSTEP.tsv

Chaosbelow:
Applied my previous changes from Lines 2136 - 2261 in a previous pull since the files have been changed. Will delete previous pull. 

Also applied a few changes I noticed when going back. As well as fixing even more Lines till 2297. 

2192 - added through based on sentence context

2196 - based on it being an item I properly capitalized it
2217 - added commence but not to sure about using the verb
2241 - changed from "to" to "for"
2247 - changed thinks to considers valuable to further emphasize what Peltastas find "most important"
2253 - not sure what "pull" is intended to convey here, so changed it to use
2255 - improved sentence flow 
2257 - maybe already implies that the monsters could have them, removing may because it already is implied with maybe.
2261 - changed Mugosa to Wugushi based on the latter text mentioning poisons. cant seem to find 우구시 in the korean version of the same line. also shown in issue #299 
2266 - changed to "nothing" based on the implication that Hackapell master thinks EVERYTHING can be penetrated 
2275 - changed to will to keep future tense meaning
2278 - changed the clause "don't you think" to the end of the sentence to improve sentence flow and question function
2283 - not entirely sure about this line, is the enemy rushing in at you? or are you rushing at the enemy? added one small change for now
2285 - not too sure what the intended meaning is here, just added on for now
2296 add 's to apply possession of her own magic

### DIFF
--- a/QUEST_JOBSTEP.tsv
+++ b/QUEST_JOBSTEP.tsv
@@ -2133,18 +2133,18 @@ QUEST_JOBSTEP_20150317_002132	Defeated Honeypins. Report to Corsair Master.
 QUEST_JOBSTEP_20150317_002133	Talk to Doppelsoeldner Master
 QUEST_JOBSTEP_20150317_002134	
 QUEST_JOBSTEP_20150317_002135	
-QUEST_JOBSTEP_20150317_002136	Report to Doppelsoeldner master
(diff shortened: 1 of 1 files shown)


imcgames:
Thank you :smile:

PR #453, opened 2015-07-05
Update ITEM.tsv

vegax87:
Added and corrected a lot of potions, stamina pills and EXP Cards


Sunaries:
My personal opinion is that I like "Lv" more than "LV". It's easier and distinguish on the eye.

vegax87:
I understand, but IMCGames used uppercase for item levels during CBT2, so I stick with it

imcgames:
if Lv is better, we could change to Lv actually :wink:

Sunaries:
@imcgames :thumbsup:

PR #454, opened 2015-07-05
Update ITEM.tsv

Sourpusss:
Line 3229 is a google translation (Korean -> English worked best, this time) with editing based on context clues.  The automated translation mentioned a canyon, so I assume it's referring to Pillar Canyon.  The first word was "decisions," so I changed it to Decidedly.

It might be mentioning Pagasa Cliff.  Perhaps this line shouldn't be added, due to the bad translation.

For reference the original line was: 부러진 기둥 협곡의 봉인을 해제하는데 필요한 결정석

This is up to interpretation / requires assistance.


imcgames:
thank you :+1:

PR #455, opened 2015-07-06
Update QUEST_LV_0100.tsv

Sourpusss:
Lines 84 and 85 has a discrepancy between the spelling of "Duckey."  The monster in the tosbase database is also listed as Duckey, so I changed all further instances to that spelling.

I made the instance of Sauga plural by putting an s on the end, since that's how it is in the line below.


imcgames:
Thank you :+1:

PR #456, opened 2015-07-06
Update ITEM.tsv

emailboxu:
Filling in lines. Editing some old lines.

Changed all instances of "Coarse" to "Crude" in weapons section, because it makes more sense in terms of equipment.

~700


imcgames:
Thank you :smiley:

PR #460, opened 2015-07-08
Update QUEST_JOBSTEP.tsv

Chaosbelow:
Proofread and edited lines 330 - 432

Still not sure whether to add "the" in front of most instances of a job master. Like "the Highlander Master" instead of "Highlander Master." Left it alone for now. 

344 - changed the wording a bit
347 - made sentence flow better
367 - changed the wording a bit
414 - not sure about this line, needs better translation
420 - need better translation
425 - changed it up a bit cant tell if this is accurate

A lot of these lines need better translation. I stopped for now before I try and retranslate or rearrange the sentences.


imcgames:
Thank you :smile:

PR #462, opened 2015-07-09
Update QUEST_LV_0200.tsv

Flevalt:
Changes:

Line5: aquire -> acquire (grammar)
L8: would need to -> need to (logic?)
L29: corvering with -> covered with (grammar)
L29: Where was the Goddess? -> Where is the Goddess? (tempus)
L31: Never would be eaten until Goddess come back -> Never would I be eaten until the Goddess comes back. (missing word | time)
L34: Are we exist -> Do we exist (grammar)
L37: do you want -> you want (logic?)
L37: more better -> even better (comparative)
L47: Ishisineti cliff -> Ishisineti Cliff (name capitalization)
L48: Ishisineti cliff -> Ishisineti Cliff (name capitalization)
L59 looks likes -> looks like
L74: soldiers is suffering the pain -> soldiers suffering pain (grammar)
L78: to the Commander Julian -> to Commander Julian (grammar)
L83: I've seen a lot so long time -> I've seen a lot in all this time(sense?)
L87: the unperturbed principles -> unperturbed principles (logic?)
L89: is -> have (grammar)
L90: as no one here -> as no one is here (missing word)

Yey or Nay?

### DIFF
--- a/QUEST_LV_0200.tsv
+++ b/QUEST_LV_0200.tsv
@@ -2,10 +2,10 @@
 QUEST_LV_0200_20150317_000002	The monsters' attacks are becoming more violent. {nl} I'm determined to succeed, though I fear for our safety.
 QUEST_LV_0200_20150317_000003	I don't care if I give up my life today if I will become a dead man tomorrow.
 QUEST_LV_0200_20150317_000004	Commander Vacenin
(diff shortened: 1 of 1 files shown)


imcgames:
Thank you :smiley:

PR #463, opened 2015-07-09
Update QUEST.tsv

megaoka:
Translating again!

### DIFF
--- a/QUEST.tsv
+++ b/QUEST.tsv
@@ -29,7 +29,7 @@ QUEST_20150317_000028	William
 QUEST_20150317_000029	The fee's 100 Silvers. {nl}It'll be cheaper than cutting through monster central.
 QUEST_20150317_000030	Cane Bug
 QUEST_20150317_000031	I charge 100 Silvers. {nl}That will cover operations, labor and other stuff.
-QUEST_20150317_000032	Hey, wait a minute. {nl}You know you need more than that...
+QUEST_20150317_000032	Hey, wait a minute. {nl}You know you need more than that…
 QUEST_20150317_000033	That's no good! {nl}I can't do something with nothing, even if it's to help the Revelator.
 QUEST_20150317_000034	Wandering Merchant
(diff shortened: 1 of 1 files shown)


ensata:
Um, are you translating or are you editing? I re-translated lines 1-150 last weekend, and I think you made a few lines inaccurate by adding words.

The unicode ellipsis was added by me. There's a technical reason for why I recommend using them, but it wasn't discussed properly last week.

Sourpusss:
Don't change ellipsis, please.  Literally all ellipsis are ".. " within the database.  It's best to just leave them how they are, than start changing them and miss some.

megaoka:
You can change all the elipsis back, but they're a large majority "..." not "…"

it's all over the place. Wouldn't it be better to change it to be uniform than just not change it at all?

@ensata, translating and editing. If you disagree, I do not mind changing them back. If there's a technical reason for using unicode ellipsis, it's a very easy change to fix all the types of ellipsis we have on the project.

ensata:
Actually, the majority is currently "..", and no one seems to want to fix or standardise it, even though it's incorrect grammar. But we shouldn't discuss that here.

Anyway.
In line 37, the second sentence should be a singular 'companion', but I prefer the original.
In line 77, Henrica never says 'Welcome', but that's a bit frivolous. Whatever.

megaoka:
Yeah, I meant ".." not "..." just coming back to the project after surgery so I'm a little out of the loop right now.

I've changed line 77 back. I'll find another section to work on if you are going through this one methodically. I don't mean to redo work that's already been done.

Sourpusss:
It's fine, lol.  But yeah, for simplicity's sake and uniform in the incorrect, just leave them.  Technically, ellipsis shouldn't be used that way in the first place.  It's meant for omission, not pause.  But this is an online game, not a thesis.

imcgames:
Thank you and welcome back  :grinning: 
I hope you're alright now :pill: :syringe:

PR #472, opened 2015-07-12
Update ITEM.tsv

Sourpusss:
Just a few edits

Line 3417 might need a [nl], but I'm not sure what the character limit is for the text box.


imcgames:
Thank you :+1:

PR #482, opened 2015-07-13
Update QUEST_LV_0100.tsv

Pomez:


Sourpusss:
You should revert all the ellipses back to the original ".. " (two periods and a space) format.  This is how the majority of them are, and until we get some input from IMC, I think we should keep them that way.

imcgames:
Thank you :smile:

Pomez:
@Sourpusss Ah i see, sorry I didn't notice it. Thanks for your reminder :smile:

PR #490, opened 2015-07-15
Update ITEM.tsv

Flevalt:


imcgames:
Thank you :smile: 
(seems to have contained machine translated lines which were wrongly translated.
We'd like to recommend you not to use google translators for long lines. :wink:  )

PR #491, opened 2015-07-15
Update ETC.tsv

LevyAM:
Just a few adds at the end of the file


Sourpusss:
The translations seem good, but you should fix your spacing to be uniform with your other lines, as well as the rest of the files.  Also, you should add x's in front of your edited lines (see the issues thread for more information, if needed).

imcgames:
Thank you :+1:

PR #492, opened 2015-07-15
Update Lines 0 to 705 from ETC.tsv English revision

ttgmichael:
As requested, here is a new pull request where my repository is up to date with your repository. Changes are also labeled with 'x' in the front of "ETC". I spent some more time translating lines so I got through to lines 705. I seldom changed titles/names and only mostly altered the descriptions of monsters in the file.

Let me know if you have any other requests! :)


imcgames:
Thank you :smile:

PR #495, opened 2015-07-15
Update QUEST.tsv

Sourpusss:
Marked all edited / proofed lines with an "x." Lines needing further proofing / proper translating are listed below:

Line 217 needs a better translation.  The current line is Kr->Jp->En, while Kr->En translates to "Robbers broke Lou's"  Whatever the case, I didn't mark it as completed.  I think the proper translation is something like "Robber Lou's Report."


imcgames:
Thank you :smile:

PR #498, opened 2015-07-16
Update ITEM.tsv

emailboxu:
Sorry I've been MIA, work's been hella busy..

Starting from the bottom going up.

Need confirmation on 5508. "Aquatic herb" or just "plant" seemed wrong, so I used a less common word for it. Not sure what the 아실라도 is though, closest I could find is "acylation", which doesn't quite seem right. 

5498, not sure what 우드르와 is. Made it up. Also in 5493, 5494, 5499.

5489~


imcgames:
Welcome back and thank you very much! :grinning:

PR #500, opened 2015-07-17
Update QUEST_JOBSTEP.tsv

Pomez:
Just edited some typos and names. I tried "translating" some missing lines but gave up since it was too frustrating.


emailboxu:
I'm pretty sure Evoniphon is his English name, there's no phonetic equivalent to the "f" sound in Korean so it comes out sounding like a "p".

Pomez:
Ah i see, since I checked ETC file and it said "Evonipon". I'll edit them now. Thank you :smile:

imcgames:
Thank you :+1: 

(we will keep Roxona Marketplace as is :wink: )

PR #502, opened 2015-07-17
Update QUEST_LV_0200.tsv

Flevalt:
Lots of typos.

2 lines with the word "sould" popped up. 
In the 1st line it made more sense to replace it with "sound", 
In the 2nd line "soul" made more sense.


imcgames:
Thank you! :smile:

PR #503, opened 2015-07-17
Update ETC.tsv

Sunaries:
Why is NXP referred as 'Coins'? Let's stick to the original currency before we make up new terms for the virtual currency. I haven't seen an Issue on Github about this. I updated some sentences since some of them could be much better.


TOSBase:
It was changed to Coin by imcGames ;) (NXP doesn't really makes sense for the international version)

https://github.com/Treeofsavior/EnglishTranslation/commit/b8fd5e88fc51b0df80365e06e2334d0a3d69d03b

Sunaries:
@TOSBase Thanks for the clarification. I did not know that. I will change it to Coin. 
However, isn't there a better term than Coin to indicate a virtual currency in the International server?

Sourpusss:
Technically, Gold Coins are the term for NXP and Silver [Coins] are used for 100% digital currency.

TOSBase:
In the wiki it's currently referenced as
도펠죌트너 Doppelsoeldner
https://github.com/Treeofsavior/EnglishTranslation/wiki/2.1.%20Class

In the commit there are some "Doppelsodner" and a "Doppelsoldner"

Sourpusss:
Technically it should be Doppelsöldner

https://en.wikipedia.org/wiki/Doppels%C3%B6ldner

ETC.tsv
```diff
@@ -2629,15 +2629,15 @@ ETC_20150317_002628	My Page
 ETC_20150317_002629	Last Visit
 ETC_20150317_002630	Open board and logout
 ETC_20150317_002631	Coin cycle
-ETC_20150317_002632	No Coin you can give.
-ETC_20150317_002633	Please enter the number of Coins shown.
+ETC_20150317_002632	You do not have any Coin to give.
+ETC_20150317_002633	Enter the amount of Coin you want to gift.
```
emailboxu, line 2633: "number of Coins" is better here IMO.

ETC.tsv
```diff
@@ -2629,15 +2629,15 @@ ETC_20150317_002628	My Page
...
 ETC_20150317_002634	{Auto_2} ~ {Auto_1} ea
-ETC_20150317_002635	I received {Count} Coins as a gift from {Giver}.
-ETC_20150317_002636	I gifted {Count} Coins to {Receiver}.
-ETC_20150317_002637	{Giver} Mr. Gift Coins is missing.
-ETC_20150317_002638	{Receiver} can no longer receive Coins for today.
+ETC_20150317_002635	You received {Count} Coin as a gift from {Giver}.
```
emailboxu, line 2635: Personally I think it should be "Coin(s)". The plural sounds better for more than one.

imcgames:
Thank you :+1: 

(NXP was changed to Coin but we are going to change it to iCoin :wink: )

PR #514, opened 2015-07-20
Updated ETC.tsv File

Kxizen:
Added Translations. 6704-6708


imcgames:
These seem to be wrong lines :sweat_smile:

PR #517, opened 2015-07-20
Update QUEST_LV_0200.tsv

Flevalt:
Minor fixes.


Flevalt:
Updated.
What do you mean with x after the edit?
Shouldn't the x be in front of the Code?

imcgames:
Thank you :smile:

PR #519, opened 2015-07-21
Update QUEST_LV_0100.tsv

Sourpusss:
The wording on line 5279 is kinda awkward.  Not sure if it's talking about the true demon (in the past) or the demon in human form. I left the "x" off for further editing.

Line 5298 refers to the previously mentioned "Chesed Altar" as "Cetek Altar."  I kept it as Chesed for uniformity.  Correct me if I'm wrong here.  Also, not sure if Sviesa Altar is a third, previously unmentioned altar, or a different translation for the previously mentioned Gedula altar. Finally, I'm not sure if a seal is on the altars, or the altars contain a beneficial seal.  I just left it as is ("seal of Sviesa Altar," versus "seal on Sviesa Altar").  I kept the "x" off for now, and left it as Sviesa.

From line 5303 on till the end of the quest, the description of the object being gathered keeps changing from "gravestone" to "memorial stone."  I made it gravestone, only.

Line 5311 mentions "unique slate," of which is implied to be under the bridge.  However, I'm not sure if it's referring to a patch of slate in the ground / in the bridge, or a chunk of slate laying around on its own??  I'm going with the idea that it's a piece that fell off underside of the bridge.

Line 5317 brings up Treeambulo, which I changed to be plural with an -es ending (like Buffalo -> Buffaloes).

Line 5320: Based on the previous formatting of the quest, the quest tracker lines were sometimes listed after the objective was completed / listed them twice (quest tracker and then some other context).  That being said, I feel it's referring the Treeambulo pieces again, but I'm not 100% that it isn't referring to a new "Carbonized Piece of Ruin."  But I don't think that's correct - please provide some input.  Left the "x" off for now.

### DIFF
--- a/QUEST_LV_0100.tsv
+++ b/QUEST_LV_0100.tsv
@@ -5263,11 +5263,11 @@ QUEST_LV_0100_20150317_005262	Defeat the monster that followed Toby
 QUEST_LV_0100_20150317_005263	Todou is being chased by Yonazolem! Defeat the Yonazolem.
 QUEST_LV_0100_20150317_005264	Defeat Yonazolem that Todou brought along
 xQUEST_LV_0100_20150317_005265	Rexipher is thinking deeply, looking at the clues. Talk to him.
-xQUEST_LV_0100_20150317_005266	Get Hogma's Teeth
+xQUEST_LV_0100_20150317_005266	Get Hogma Teeth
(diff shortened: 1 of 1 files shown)


imcgames:
Thank you :+1: 

-5279 refers to the past
-For 5298, Cetek altar would be the correct one and Sviesa is correctly translated. 
-Yup, we can go with gravestone as it's shorter :grin: 
-well it refers to some pieces/debris/fragment from the ruins which were carbonized

PR #524, opened 2015-07-21
Update ITEM.tsv

emailboxu:
5465 - ? Not sure what 뚜벅초 is.

5437~


imcgames:
Thank you :smile: 

뚜벅초 is a grass used as ingredient to make a bomb

PR #525, opened 2015-07-21
QUEST_LV_0100

froodybanana:
- removed a bunch of {nl} from the end of sentences, assuming the game
  engine takes care of proper spacing between paragraphs
- removed a bunch of whitespaces around {nl}.
- updated a number of lines/sentences between 1-250

QUEST_LV_0100_20150317_000113 - previous line mentions assistant named
Irene and then uses "he" to refer to said assistant. Irene being a
womans name I changed this to she instead. Irene is also later referred
to as "her".  but without context it's hard to know if this line is
actually referring to Irene at all. could be someone else entirely.
QUEST_LV_0100_20150317_000173 - changed Greak to Great as Greak isn't a
word and it doesn't seem to appear anywhere else in the document, so I'm assuming it's a typo

### DIFF
--- a/QUEST_LV_0100.tsv
+++ b/QUEST_LV_0100.tsv
@@ -230,7 +230,7 @@ QUEST_LV_0100_20150317_000229	So you finally came.{nl}What did the Bokor Master 
 QUEST_LV_0100_20150317_000230	I'm also curious about the girl who cracked the seal.{nl}This is not a seal that can easily be broken.
 QUEST_LV_0100_20150317_000231	If the Bokor Master told you that I would know it..{nl}I can think of a place.
 QUEST_LV_0100_20150317_000232	
-QUEST_LV_0100_20150317_000233	Miner's Village Mayor
+QUEST_LV_0100_20150317_000233	Miners' Village Mayor
 QUEST_LV_0100_20150317_000234	Welcome. Aren't you our town's savior?
 QUEST_LV_0100_20150317_000235	
 QUEST_LV_0100_20150317_000236	So this revelation is Goddess Laima and we have to collect all these for the Goddess to regain her powers?{nl}That is amazing.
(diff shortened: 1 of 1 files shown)


vtange:
So the {nl}'s at the end of lines are not needed? I always left them there thinking the game might need them to set up for text that comes after

ttgmichael:
They are needed. Refer to the wiki: 

The tag '{nl}' starts the translations on a new line. If the translations of a line is too long, you may use {nl} to change lines where necessary. 

This should means that the game will not take care of the new lines. You must specify the new lines for the text to fit within the UI.

froodybanana:
The way I read it every line starts a new paragraph. Every {nl} cuts the current paragraph. Thus having {nl} at the end of a paragraph makes no sense, because there's no reason to cut the line when nothing follows.

For example:

QUEST_LV_0100_20150317_000001   This is a very long sentence that is being{nl}cut in half so it fits on screen.
QUEST_LV_0100_20150317_000002   This is another very long sentence but it's in a new paragraph.
QUEST_LV_0100_20150317_000003   Ending with nl doesn't make sense{nl}

Both from a programming point of view and a writing point of view, ending a paragraph with a new line doesn't make sense.

Sourpusss:
It makes things look cleaner.
Things will appear to flow more with the characters speech.

I don't agree with this choice, and you're more likely to run into ugly word wra
pping when the dialogue windows are too small to hold these long sentences
you've added.

(I hope you can see my illustrations of the points).

Sourpusss:
You need to be careful, as you're more likely to run into ugly word wra
pping when the dialogue windows are too small to hold these long sen
tences you've created (I'm not sure what the character limit is).

It might look a bit weird sometimes, but for the most part,
it shouldn't be that big of a deal.
As long as you place them after some form of punctuation,
it more or less seems like it's flowing with the characters' speech.
It's almost poetic in appearance.

(Hopefully you can see the points I'm illustrating)

imcgames:
Some lines may be used together in one dialogue box so I think it'd be safer to leave the {nl} 's at the end of the lines if the Korean (original) text has it. 
Would it be possible to resend us the pull request without removing the {nl}'s at the end of lines?

froodybanana:
Prrrr... yeah I suppose. Let me see what I can hack together.

PR #531, opened 2015-07-22
commit ea459e9 revisions (skip the 24 hour wait thing?)

Sourpusss:
Various corrections based on imc's input regarding my previous merge.

One thing: I'm guessing Line 5279 is a quest tracker line? I made it refer to the past like you said, but I wanted to make sure there aren't too many characters for its UI position.


imcgames:
Thank you :grinning:

PR #533, opened 2015-07-22
Update QUEST_LV_0100.tsv

froodybanana:
- Removed many unnecessary whitespaces
- Updated dialogue in lines 001 - 510
- Replaced Miner's Village with Miners' Village (unless there really
  only is one miner and it's his village)

QUEST_LV_0100_20150317_000113 - previous line mentions assistant named
Irene and then uses "he" to refer to said assistant. Irene being a
womans name I changed this to she instead. Irene is also later referred
to as "her". but without context it's hard to know if this line is
actually referring to Irene at all. could be someone else entirely.
QUEST_LV_0100_20150317_000173 - changed Greak to Great as Greak isn't a
word and it doesn't seem to appear anywhere else in the document, so I'm
assuming it's a typo
QUEST_LV_0100_20150317_000255 - not sure what to make of this line
really
QUEST_LV_0100_20150317_000278 - I am not sure what to make of this bit
{nl}{-scp
SCR_DIALOG_NPC_ANIM( .. does it need to be there or not?
QUEST_LV_0100_20150317_000279 - a number of lines including this one
start with the symbols )}.. is there a reason for this?

### DIFF
--- a/QUEST_LV_0100.tsv
+++ b/QUEST_LV_0100.tsv
@@ -1,117 +1,117 @@
 ﻿QUEST_LV_0100_20150317_000001	Sentinel
-QUEST_LV_0100_20150317_000002	You better go straight to Shadow Forest Road if you want to go to Klaipeda. {nl} It is dangerous in the deep woods so try to stay away if you can.
+QUEST_LV_0100_20150317_000002	You better go directly via Shadow Forest Road if you want to go to Klaipeda.{nl}The deep woods are dangerous, so try to stay away if you can.
 QUEST_LV_0100_20150317_000003	Man in hiding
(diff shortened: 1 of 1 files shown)


froodybanana:
Managed to re-introduce all the removed {nl} instances somehow. Still I wish they didn't have to be there. It's like an itch I'm not allowed to scratch...

imcgames:
The updates were made only on 001 - 510, right? 
There seems to be some conflicts so we'll only merge lines 001 - 510 then :smile: 

(QUEST_LV_0100_20150317_000255 - fixed it
QUEST_LV_0100_20150317_000278 and others like this, it's part of how the scripts go in game so if the Korean text has it, then English should have it as well. :wink: )

PR #546, opened 2015-07-23
Update ITEM.tsv

emailboxu:
5394 - I can't think of a better word for 눅진한 기름, it's one of those "not really a good translation to English" words.

5396 - 아이트바라스 = ? and 나뮤 사원터 = ? Took the translation from line 5385.

5388 - I based the translation on the line below it.

5372 - The rubbing meaning is ambiguous and require context. 

5374 - Changed to "silver spoon", as that's the English version of the idiom.

5352 - Simplified to "Monster Leather". Correct me if wrong.

Not sure if you wanted "monument" or "tombstone." Another thing that will need context. Currently just left it as tombstone.


ensata:
For 5394, try 'Pulpy Oil', 'Mushy Oil', or 'Gooey Oil'.

imcgames:
Thank you :smile:

PR #548, opened 2015-07-23
QUEST_Lv200.tsv - Stone Frost fix

Flevalt:
All petrified frosts and stone frosts have been changed to the correct form Stone Frosts.


imcgames:
Thank you :smile:

PR #557, opened 2015-07-25
error correction

ensata:
minor errors that imc introduced in merging.

56 - since someone didn't agree with my attempt to make it sound colloquial and confused, I've gone back and based it on the pre-edited version. (The actual translation is something like 'thin, scrawny, emaciated and ill-looking')

185 - this is from the Wiki.

200 - This is the proper translation, it's an actual Lithuanian name, and you added this spelling yourself to etc.tsv


imcgames:
Thank you :smile:

PR #558, opened 2015-07-25
Update ITEM.tsv

Sourpusss:
Edit and Review of lines 2716 through  2778

---

Lines 2761, 2765, 2768, and 2770 discuss stones referred to as "Pyrite" and "Cryorite." For pyrite, the english meaning of the stone also  refers to "Fool's Gold." I'm not exactly sure if this is meant to be the case, or if calling it "Pyroite" would be better suited.  For Cryorite, there is a real stone called "Cryolite."  As with "Pyrite," I'm not sure if we're referring to real (magically enhanced) or purely magical stones.  This being said, I left the terms untouched and the lines without x's for now.  Of the various options to go with, I feel like we have:

"Pyrorite" and "Cryorite" (Most fitting in magical terms, but least tied to real-world stones) [My vote]
"Pyrolite" and "Cryolite" (Also fitting, but both stones have real world meanings [Pyrolite is a theoretical stone in the earth's mantle])
"Pyrite" and "Cryorite" (Everything left as is)
"Pyrite" and "Cryolite" (Most realistic - all stones are the names of actual minerals in the english language)

### DIFF
--- a/ITEM.tsv
+++ b/ITEM.tsv
@@ -2713,69 +2713,69 @@ ITEM_20150317_002712	Devilglove Fragment
 ITEM_20150317_002713	 
 ITEM_20150317_002714	Moa Feather 
 ITEM_20150317_002715	 
-ITEM_20150317_002716	Parchment 
-ITEM_20150317_002717	Luxurious parchment. You can write anything on it. Purchasable from the Pardoner Master. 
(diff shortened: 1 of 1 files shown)


imcgames:
Thank you :smile:

PR #561, opened 2015-07-27
Update QUEST_LV_0100.tsv

Sourpusss:
Editing and Proofing of Lines 3000-3050

There are several lines, such as lines 3005, 3030, 3035, 3039, and 3050  in which the text seems to be some kind of script? I left them untouched and without x's.

Line 3009: I'm not sure if this is the player reading a book and discussing it internally, or some kind of first-person quest narration? I left it untouched.

Line 3019: "the tower of wizard" is a frequently used translation of what I believe should be Magician's Tower.  The term "Magician's Tower" is used a few lines down, and it would make sense that these are both the same location.  I'll bring this up in the issues, shortly.

Line 3020: I'm not sure what this should be?  I don't know if they mean a Reject in terms of a noun (rejected person), or the verb in some context. I left it alone.

Line 3027: Just a small edit, but I added the period there and noticed there was a space after "way," so I added a space after the new period.


imcgames:
Thank you :smile:

PR #568, opened 2015-07-28
Patch 1

magcathe:


ttgmichael:
Rather than being a pull request, this should be a separate repository all together. This repository is for Korean to English translation (hence the name). It is up to IMC to decide whether to create an official French repository.

In the meantime, it's great that you have the initiative to get the French translation started. Just make sure you keep updated with changes made from this repository. As long as you do what you are doing at a separate repository, there might be a French localization sooner than if you don't do this.

imcgames:
Unfortunately, these files seem to be old versions and we can't merge it if it contains French as mentioned above.

PR #570, opened 2015-07-28
QUEST 1000 - 1102 + Ellipsis Fix

vtange:
Will go back through the 1000s for {nl}'s since they are very long lines.


imcgames:
Thank you :+1: 

(The ellipsis seemed alright so we left it as is for now)

PR #577, opened 2015-07-29
quest.tsv - 251-300 + a etc&item supp

ensata:
quest.tsv - batch 251-300
tl check, retranslate, rewrite
Some supplementary lines in etc.tsv and item.tsv were also changed.
Needs further work for keyword ref synchro


imcgames:
Thank you :+1:

PR #578, opened 2015-07-29
Line revision 9900 to 10000+, 12350-12450, some other lines, Replaced Faint with Stunned, Magic field with Magic circle, other minor consistency changes

ttgmichael:
Title says it all. 

Revised/reviewed lines 9900 to 10000 plus lines here and there for consistency I noticed both stunned and faint being used, changed to stunned (best word for "to make faint" that I can think of in English). Both Magic field and Magic circle were used, Googling the korean term seems to lead more results towards Magic circle, so I'm putting it as magic circle.

Oh and I changed Chamber of Secrets to a different name (ToS isn't Harry Potter xD)

### DIFF
--- a/ETC.tsv
+++ b/ETC.tsv
@@ -1670,7 +1670,7 @@ ETC_20150317_001669	Must be at least {LEN} characters.
 ETC_20150317_001670	Must be less than {LEN} characters.
 ETC_20150317_001671	Contains one or more characters that are not allowed.
 ETC_20150317_001672	You cannot ride during Rest Mode.
-ETC_20150317_001673	You cannot ride on a fainted companion.
+ETC_20150317_001673	You cannot ride on a stunned companion.
 ETC_20150317_001674	The customer is not a CBT user.
 ETC_20150317_001675	It has been banned by the administrator.
 ETC_20150317_001676	{Name} ({Mapname} / channel {ChNo})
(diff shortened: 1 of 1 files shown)


vegax87:
Fainted is a term to describe a KO, a collapsed person or animal, and not a stun, when your mount dies you cannot ride it, of couse...

ttgmichael:
I'll happily change it all to faint then? I am hesitant on thinking faint == KO like in Pokemon because of some lines:

ETC_20150317_002282 had its translation as:
You've been stunned by the root's poison and fainted!

Would the poison KO you? It felt more appropriate for stun.

ETC_20150323_009981 had its translation as:
Stun Scorpio to faint and extract poison

As least from the gameplay videos I watched, monsters that are killed will explode and disappear. So it wouldn't be possible to extract poison from the Scorpio if you are to KO it. Stun felt more appropriate (to immobilize the Scorpio/ make it unconscious).

ETC_20150714_012377 had its translation as:
You've shot Wendigo with the paralyzing needle!{nl}Pull out the hairs of fainted Wendigo Archer in 10 secs!

Why would a paralyzing needle faint/KO someone? It sounded more appropriate to use stun.

What are other peoples thoughts? ^^ @imcgames would help if you can clarify "faint"/stun (Korean term is 기절시키고 ), that'll help! :)

ttgmichael:
Revised/reviewed lines 12352 to 12450 as well; Stun changed to Faint

imcgames:
Well.. ETC_20150317_002282, ETC_20150714_012377 is more like it's fainted already
while ETC_20150323_009981 is to make faint.
So I guess use of faint/stun would depend on the lines?

ttgmichael:
I've changed the wording so they can all be fainted :)

imcgames:
Thank you :+1:

PR #580, opened 2015-07-30
Sidequest - Hidden Treasures

ensata:
standardised the translations between quest-etc-item for the Hidden Treasures sidequest lines.
Also translated the relevant quest journal entries.

considered experimental due to cross-dependence over 4 files.


imcgames:
Thank you :smile:

PR #587, opened 2015-07-30
mapnames for cbt navigation

ensata:
etc.tsv 1140-1450 - the main block of map names
three types of edits:
1. a dozen or so map name changes based on a quick look
2. changed levels to conform to English styles, also for length (ie. 1st Floor -> 1F)
3. added letter codes to duplicate names for faster identification for English speakers; mapcodes are still in brackets and should be removed later.

(PS: due to the amount, the map name suggestions are collated here: https://github.com/ensata/ToS-en-fork/blob/notes/mapnames.tsv )


imcgames:
Thank you :smile: 

(map codes are only used in system so we'll just leave it as is. 
As for the map name changes, we've merged most of it, and changed them in the wiki as well. )

PR #590, opened 2015-07-31
fixing potential rendering bug for cbt

ensata:
mass search and replace of all {Nl} tags to {nl} to prevent a potential rendering error before the cbt
please note that this tag error may indicate that all surrounding lines came from a cut-and-pasted machine translation.

most of the incorrect {Nl} tags were in quest.tsv - lines 300-1300; others were removed in older commits

I do not normally like this type of edit, but since the cbt is approaching, I wanted to highlight the issue and give the fix at the same time - so it's up to you if you think it's actually required.


imcgames:
Hi, we'll be doing this later on so we'll close this for now :wink:

Issue #581, opened 2015-07-30
Consistency question - Lithuanian language in item names


Xenobius2501:
Hello.

So far I've been mostly reading the forums, but recently one thing caught my attention.
As you probably know by now, a large number of in-game terms (locations, NPC names, items and recipes etc.)  contain Lithuanian words. 

In case of NPCs, cities and most locations it's reasonable, although IMO, some names like Kvailas Forest (stupid forest), Akmens Ridge (stone ridge), Verkti Square/Apsimesti Crossroad ("To Cry" Square/"To Pretend" Crossroad - yes, they used infinitive form of Lithuanian verbs) are somewhat sketchy, in case of items it's inconsistent, to say the least. 

For example, many of rather common items (food or alchemy ingredients, as far as I can tell, since I haven't played the game yet) are in English - there's Onion, Garlic, Acorn, Peach and Strawberry, but some of them are in Lithuanian for no apparent reason - there's Aviete (Raspberry), Salavijas (Sage plant), Ramune (Chamomile), Medus (Honey), Dilgele (Nettle) and so on - which I personally find rather confusing.

Similarly, most of equipment is in English (Miner Hammer, Royal Mace, Ice Rod and the like) - but some item names contain seemingly random Lithuanian words - there's Sunkus Maul - which simply means "Heavy Maul", Bendras Sword ("bendras" could be very roughly translated as "common"), Saltas Staff ("saltas" means "cold") and probably the worst offenders - Zalia Leather Gloves/Boots/Pants Armor/Kite Shield - because "zalia" is Lithuanian for green, more specifically, it's a gender-specific form of "green" only used to denote "feminine" nouns, masculine form being "zalias". 

Long story short, it brings us to the question - is such inconsistent naming policy (English terms with random bits of Lithuanian here and there) a case of "working as intended" or something should be done about it? Now, I do realize there aren't exactly many Lithuanian players, it's a small country after all, but to people actually familiar with the language such things appear rather silly at best. 

As you might have guessed by now I happen to know Lithuanian language myself, and while I'm not even close to being a professional linguist I could (if necessary), given some time, point out such cases and perhaps suggest translations - that is, if leaving it the way it is wasn't the grand plan all along.

ensata:
Well. You might not be a linguist, but if you know what an infinitive is, you know more than most people.

[...]

PR #591, opened 2015-07-31
Location name changes to fit Lithuanian grammar.

ttgmichael:
Taking advice from @Xenobius2501 at
https://github.com/Treeofsavior/EnglishTranslation/issues/581

Changes to the Lithuanian location names in ETC.tsv

I'm guessing: (let me know what should it be ^^)
Tiltas Valley has more than 1 bridge. -> Tiltu
Ramstis Ridge being called Pillar Valley (singular)  -> Ramscio
Dvasia Peak as Soul's Peak (singular) -> Dvasios
Poslinkis Forest as Shifted Forest (past-participle) -> Paslinkes
Septyni Glen as 7th Glen -> Septyntas 
Vieta gorge to Vietos for grammar's sake.
Veja to Vejos too
Kvailas to Kvailio


Xenobius2501:
@ttgmichael 

Looks legit so far.

I'll try to lay the groundwork for further name fixes later today after I go through my usual workload - probably checking the remaining lines on location list comes first. Also, being new to GitHub in general - where would it be better to post suggestions - here, in the form of comments or back in #581 ?

ttgmichael:
Great!

If you want to post a suggestion for something related to the changes I've made here, you comment here. If you want to post a suggestion about something new, you should post back in #581 so it's more organized. ;)

ensata:
Since these are keywords that will appear more than once, it's good practice to simultaneously add them to a reference file, so that you won't lose track of it so easily. This pull request will close and vanish eventually.

imcgames:
Since these are actually spread out in other files, we will try to change them together with the other files later on. 
So we'll close this merge for now.

PR #592, opened 2015-07-31
Update QUEST.tsv

wlghksdl2:
Translated 2332 and 2333. 
Translated up to 2342. Need help translating the following line.
그러나 곧 앞으로 내기를 안 하면 그만이지 라는 생각이 들었고, 곧이어 해도 이기면 그만이지 라는 생각이 들자 저도 모르게 고개를 끄덕여 승낙하고 말았습니다.
Does this mean his line of thinking is "Even if I lose then all I have to do is stop gambling. Even if I gamble, despite losing means a friend will disappear, all I have to do is win anyway?" Is that why he takes on the bet?
Translated to 2350. Need help with the following.
모데스타스씨는 정말 필사적으로 마음속으로 싸웠습니다 Not sure how to translate it.
In context does it mean "Modestas desperately fought his inner urges to take on the bet?" 
But he takes on the bet anyway. Was he unsure whether this is worth his first win?
Added translations up to 2359
2356 - need help translating 
Added up to 2362
Added up to 2370
Added up to 2375
Added up to 2385. Finished (save a few sentences) the story of Modestas the Gambler


imcgames:
Thank you :+1: 

(Nopa means old lady :smile: )

PR #593, opened 2015-07-31
ETC.tsv Reviewed lines 3500 to 3651

ttgmichael:
Checked the translations on these lines and putting o's on them. Most of them were good! Would be terrible to see ETC_20150317_003619 in the CBT ^^;


imcgames:
Thank you :smile:

PR #598, opened 2015-07-31
Update UI.tsv

Sunaries:
Minor update. Hopefully OTC is available in the English CBT, so that I can adjust it accordingly. It's rather exciting to edit in-game.


imcgames:
Thank you :smile:

PR #600, opened 2015-07-31
Update SKILL.tsv

emailboxu:
Posion -> Poison
1 time(s) -> 1 time
2+ time(s) -> 2+ times
Attack area -> AoE (Area of effect, common term that should be familiar)
Fan-shaped -> Conical
Head-butt -> Headbutt
Fuem -> Fume
X hand/X-hand -> X-handed
Removed instances of triple commas 

Translated as much of the Korean stuff as I could find. Point out more if they exist.

Stopping because I'm at 269. Limit of 300 per PR, right?

### DIFF
--- a/SKILL.tsv
+++ b/SKILL.tsv
@@ -7,7 +7,7 @@ SKILL_20150317_000006	[Raffly] Physical, Strike, Poison Property, Jump Attack 1 
 SKILL_20150317_000007	[Puragi] Physical, Hit, Poison Property, Normal Attack 1 time
 SKILL_20150317_000008	[Puragi] Physical, Strike, Poison Property, Attack 1 time
 SKILL_20150317_000009	[Puragi] Physical, Strike, Poison Property, Stomp 1 time
-SKILL_20150317_000010	[Zigri] Normal Attack, Frontal AoE
+SKILL_20150317_000010	[Zigri] Normal Attack, Cone AoE
 SKILL_20150317_000011	[Zigri] Rotating Attack
 SKILL_20150317_000012	[Zigri] Jump and Stomping Shock
 SKILL_20150317_000013	[Fisherman] Physical, Strike, Ice Property, Strike with Fishing Rod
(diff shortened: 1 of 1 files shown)


ensata:
AoE should be written fully, since that's advanced jargon. New players won't understand what it means. You should never use acronyms. 'Fan-shaped' should be 'Conical'. It's de facto western RPG terminology. 'Fan-shaped' is a de facto eastern RPG term.

Also, the 300 rule was supposed to be a range limit, not a number limit... you're at a range of 5000. Ah, whatever. Just don't blame me if the merge gets complicated.

Flevalt:
"Stopping because I'm at 269. Limit of 300 per PR, right?"

You're at +269 −269. Making it a total of 538.
Not sure how I feel about AoE instead of Area of Effect.
I think it should be fully written out so it is comprehensible for everyone. But people would figure it out in due time anyway if they didn't already know.

emailboxu:
I don't think people will have problems with AoE. It's an extremely common abbreviation in many games. As for the range I just replaced all so I don't think it should mess with it too much.

ensata:
You shouldn't assume that. Everyone was a beginner at some point, and info blurbs are supposed to give info to everyone.

"Fan-shaped Attack Area" was perfectly fine, too, at least for now. It's very "eastern", but you insisted that "Swordsman" made perfect sense, too. A reference file for translating the game's game mechanic terminology doesn't exist yet, either.

ghost:
I agree that "fan-shaped" or  "cone-shaped" would've been an easier to understand term for younger players and less English savvy players. I've used the word 'Conical' maybe four times in my life. "Area of effect" should stay as well if it turns out it'll fit just fine in the UI. Making things easier for beginners should take priority over making things convenient for veterans.

And yes you kind of went over the range, no one (Or at least I wouldn't) can really touch lines 2 through 4264 now because of potential merge issues.

imcgames:
Thank you :smile: 

(We'll try to have AoE back to Area of Effect later but I'm afraid it may be too long when for UI. 
Let's see during how it goes during Beta.)

emailboxu:
The length was the exact reason I didn't put it in. If there's room for it then I don't mind switching the two later.

PR #603, opened 2015-07-31
Update QUEST.tsv

wlghksdl2:
Changed Aiste to Aistė


imcgames:
I think it's better to keep it as Aiste so we'll close this pull for now :smile:

PR #610, opened 2015-08-01
QUEST #1103 - #1126 + full elipsis. 

vtange:
Running into a wall of machine translated text, like 3 to 4 sentences avg per line. Chipping away at it, pardon me if I've been working slow compared to before.


vtange:
Maxed out on QUEST tsv. My bad. it's really just the new lines and ellipsis fix.

As a cooldown from doing so many long lines on QUEST. I've treated my self to some UI.tsv fun. First ~300 lines checked

imcgames:
Thank you :smile:

Issue #526, opened 2015-07-21
How do {nl}'s work? + {nl} at the end of lines


vtange:
Hi, I spent a good chunk of last night reviewing the first 700 lines of QUEST.tsv and adding {nl}'s where I feel the line went too long without one, I put them mostly at the end of sentences and phrases. 

There were a LOT of {nl}'s though, at the end of lines. I left them alone since I thought the game might need them to set up for a line that comes after said line. For example:

"Line 1 Line 1 Line 1 Line 1 Line 1" {nl} -> Line 1 moves up for Line 2 to show up.

Is this how things are supposed to work? Or am I just being too careful and the {nl}s at the end of lines aren't needed?

imcgames:
{nl} 's starts the translations on a new line. Some lines may be used together in one dialogue box so it'd be safer to leave the {nl} 's at the end of the lines if the Korean (original) text has it.

PR #611, opened 2015-08-01
Update QUEST.tsv

Torrel:
Edited lines
96, 95, 98, 116, 120, 130, 138, 139

Just some minor fixes and changes here and there either to make it "flow" better, shorten, or to make a bit more sense.

### DIFF
--- a/QUEST.tsv
+++ b/QUEST.tsv
@@ -92,10 +92,10 @@ xQUEST_20150317_000091	And yet, it doesn't matter. Come out and face me, you mon
 xQUEST_20150317_000092	Just now, I saw a goddess appear in my dream… {nl}How surreal…
 xQUEST_20150317_000093	Still. I have a duty to defend the kingdom. It's why I'm here now.
 xQUEST_20150317_000094	Uska's an old friend of mine.
-xQUEST_20150317_000095	We did our tests together, and Uska got to where he is now with my support. {nl}Oh, those were some crazy days… It'd be nice to get together again!
-xQUEST_20150317_000096	Just don't believe some of the things he'll say. {nl}His hometown knew him for being quite the show-off.
+xQUEST_20150317_000095	We did our tests together, and Uska only got to where he is now with my help. {nl}Oh, those were some crazy days… It'd be nice to get together again!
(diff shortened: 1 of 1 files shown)


QUEST.tsv
```diff
@@ -92,10 +92,10 @@ xQUEST_20150317_000091	And yet, it doesn't matter. Come out and face me, you mon
 xQUEST_20150317_000092	Just now, I saw a goddess appear in my dream… {nl}How surreal…
 xQUEST_20150317_000093	Still. I have a duty to defend the kingdom. It's why I'm here now.
 xQUEST_20150317_000094	Uska's an old friend of mine.
-xQUEST_20150317_000095	We did our tests together, and Uska got to where he is now with my support. {nl}Oh, those were some crazy days… It'd be nice to get together again!
-xQUEST_20150317_000096	Just don't believe some of the things he'll say. {nl}His hometown knew him for being quite the show-off.
+xQUEST_20150317_000095	We did our tests together, and Uska only got to where he is now with my help. {nl}Oh, those were some crazy days… It'd be nice to get together again!
```
Chaosbelow, line 95: Remove only (It is not implied that its "only" because of he/she's help.)
Change help back to support

QUEST.tsv
```diff
@@ -92,10 +92,10 @@ xQUEST_20150317_000091	And yet, it doesn't matter. Come out and face me, you mon
...
+xQUEST_20150317_000096	Just don't believe some of the things he says. {nl}He had a reputation in his hometown for being quite the show-off.
 xQUEST_20150317_000097	Me? I find it funny, so I just leave him alone. 
-xQUEST_20150317_000098	Uh, wait. I'll take back what I said about not believing a few things.
+xQUEST_20150317_000098	Uh, wait {nl}I take back what I said about not believing a only a few things.
```
Chaosbelow, line 98: The original Korean line has the {nl} tag at the end of the line after the period. Try not to randomly place {nl} tags because we don't know how it will look in-game. https://github.com/Treeofsavior/EnglishTranslation/issues/526

The end of the sentence is not good. "I take back what I said about not believing a few things." Is what we should use for now.

imcgames:
Thank you :+1: 

(edits based on comments were applied)

PR #619, opened 2015-08-01
Update SKILL.tsv

Rezken:
- Cosmetic skill name changes like "Ice bolt" -> "Ice Bolt" and more.
- Adding forgotten dots at the end of sentences.
- Fixing some false grammar.


imcgames:
Thank you :+1:

PR #620, opened 2015-08-01
Update UI.tsv

merrun:
look weird in game 
example 
without change: Learning xxxx (#level) 
with change: Learning xxxx (level #)


imcgames:
:+1:

PR #623, opened 2015-08-01
Update SKILL.tsv

merrun:
changing description for Leather mastery which somehow uses Plate mastery's description

### DIFF
--- a/SKILL.tsv
+++ b/SKILL.tsv
@@ -4736,7 +4736,7 @@ SKILL_20150717_004735	Magic defense increase
 SKILL_20150717_004736	Magic defense increase, Physical defense increase
 SKILL_20150717_004737	Movement speed decrease
 SKILL_20150717_004738	Max SP and Magic DEF increase based on amount of [Cloth] armor equipped and Attribute level (Same weapon types from 3 ~).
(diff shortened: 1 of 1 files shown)


SKILL.tsv
```diff
@@ -4736,7 +4736,7 @@ SKILL_20150717_004735	Magic defense increase
 SKILL_20150717_004736	Magic defense increase, Physical defense increase
 SKILL_20150717_004737	Movement speed decrease
 SKILL_20150717_004738	Max SP and Magic DEF increase based on amount of [Cloth] armor equipped and Attribute level (Same weapon types from 3 ~).
-SKILL_20150717_004739	Max HP and Physical DEF increase based on amount of [Plate] armor equipped and Attribute level. (Same weapon types from 3 ~).
+SKILL_20150717_004739	Evasion increases based on amount of [Leather] armor equipped and Attribute level. (Same weapon types from 3 ~).
```
wlghksdl2: I think "number of..." is more specific

merrun: Amount already are ok I think..then again scratch that. "number of" seems to fits, also that (same weapon type from 3~ ) should be (at least 3 of the same equipment type). but its too long ...anyone could shorten that out?

imcgames:
:+1:

PR #627, opened 2015-08-01
Update QUEST_LV_0200.tsv

Bunzilla:
I wasn't sure if all of the lines of text needed a period at the end of the sentence. There are many lacking any punctuation, but I assumed these were for clickable dialogue boxes.

### DIFF
--- a/QUEST_LV_0200.tsv
+++ b/QUEST_LV_0200.tsv
@@ -1194,7 +1194,7 @@ QUEST_LV_0200_20150317_001193	I'm worried about the whereabouts of the bishop.
 xQUEST_LV_0200_20150317_001194	Use Naktis' monsters... {nl}But how can we use them... {nl}Aha! We can turn into Naktis' monsters from the sanctuary and make the other monsters destroy it.
 QUEST_LV_0200_20150317_001195	There are materials in the sanctuary with instructions on how to make a medicine that turns you into a monster. {nl}Find those materials.
 QUEST_LV_0200_20150317_001196	Now it's time to tell Naktis' monsters where Mayvern's device is located.
-QUEST_LV_0200_20150317_001197	I fell while holding the materials, and they fell under that device. {nl}It keeps moving up and down making it difficult to get them out.
+QUEST_LV_0200_20150317_001197	I fell while holding the materials, and they slid under that device. {nl}It keeps moving up and down making it difficult to get them out.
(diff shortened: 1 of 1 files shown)


QUEST_LV_0200.tsv
```diff
@@ -1186,15 +1186,15 @@ QUEST_LV_0200_20150317_001185	I feel hungry no matter how much I eat..{nl}Hey..
...
-xQUEST_LV_0200_20150317_001194	Use Naktis' monsters.. {nl}But how can we use them.. {nl}Aha! We can turn into Naktis' monsters from the sanctuary and make the other monsters destroy it.
-QUEST_LV_0200_20150317_001195	There are materials in the sanctuary with instructions on how to make medicine that turns you into a monster. {nl}Find those materials.
+xQUEST_LV_0200_20150317_001194	Use Naktis' monsters... {nl}But how can we use them... {nl}Aha! We can turn into Naktis' monsters from the sanctuary and make the other monsters destroy it.
+QUEST_LV_0200_20150317_001195	There are materials in the sanctuary with instructions on how to make a medicine that turns you into a monster. {nl}Find those materials.
 QUEST_LV_0200_20150317_001196	Now it's time to tell Naktis' monsters where Mayvern's device is located.
-QUEST_LV_0200_20150317_001197	I fell while holding the materials and it went under that device. {nl}That thing keeps moving up and down making it difficult to get it out.
+QUEST_LV_0200_20150317_001197	I fell while holding the materials, and they fell under that device. {nl}It keeps moving up and down making it difficult to get them out.
```
Chaosbelow: You don't need another comma here since "and" joins the two clauses together. If "and" wasn't there then you could use a comma.

Bunzilla: You always need to use a comma when you have two complex clauses joined together. Or, more simply, you cannot join two complete sentences without a comma. My word choice of "fell" being used twice is questionable. Since "materials" is a plural noun, "it went" is incorrect. Slid may work better but it doesn't quite fit the sentence either. Additionally, you can't use a semi colon here because the beta complex clause is dependent on the alpha clause for clarity--they are sequential.

Chaosbelow: Ah my mistake then. I am a little lacking when it comes to my English as of late ><

imcgames:
:+1:

PR #633, opened 2015-08-02
update ETC.tsv

tanaphat:
added missing ETC (ETC_20150729_013968,ETC_20150729_014017)


imcgames:
these are not in the files yet

PR #634, opened 2015-08-02
Update UI.tsv

tanaphat:
correct spelling untradeable -> untradable 
you can also change untradable to soulbound (just leaving as untradable to preserve original text)


imcgames:
:+1:

PR #635, opened 2015-08-02
Update UI.tsv

Staryx:
spelling amsters > masters

Translation for line 282: Equipment buff, Gear Buff or Armor Buff (kr text is 장비 버프)


imcgames:
:+1:

PR #639, opened 2015-08-02
Update SKILL.tsv

RenzuMame:
Changed Non-Property to Neutral


imcgames:
we'll try to keep non-property for now :smile:

PR #641, opened 2015-08-02
Update ITEM.tsv

RenzuMame:
Changed Sabre to Saber (US English)


Chaosbelow:
Sabre is still spelled as Sabre in English. Saber is not the name of the weapon.

RenzuMame:
Is it not? I used this as reference: https://en.wikipedia.org/wiki/American_and_British_English_spelling_differences

Chaosbelow:
Well you are right that it can be spelled both ways but the more common way is Sabre. There is also really no reason to change it.

RenzuMame:
The Style Guide says to use US English, so yeah.

imcgames:
Let's keep sabre for sabers for now :smile:

PR #651, opened 2015-08-02
Update QUEST_LV_0200.tsv

Requinn:
Are we keeping ellipses as two periods ( . . ) or three ( . . . )?

Line 006566: 몽당나무 is basically a Treant. Something like "Tree Body".
Line 006573: "Pink Tree Root Moles" on the wiki look like those onion
enemies?
Line 006577: Glass Mole is apparently a Level 1 Boss according to wiki.
Line 006586: I don't trust my translation of this line.

Will finish Lines 006592 through 006599 and continue looking through it later today.


Azhurel:
I think the ellipses should be three, but that's just me.

ghost:
A lot of those ellipses aren't in the original texts and don't really go well with some of those sentences anyway (the pause is too long), so I think cutting them down should be okay.

Also @Requinn, the version of the file you edited is old and is going to cause a merge conflict. You'll want to re-sync your files so they're up to date and then put your edits back in.

Requinn:
@jalevijo oh, alright ill go update and do some more edits while I'm at it. thanks for the heads up

PR #659, opened 2015-08-02
Update UI.tsv [1-1766, all lines]

ThiefZero:
Went through the entire thing. Fixed typos and made some changes to some odd sentences. Also removed '.' and shortened other UI text where it was possible. Hard without context, hope I didn't change too many things that were intended to be that way.


imcgames:
Thank you :+1:

PR #660, opened 2015-08-02
Update ETC.tsv

DracoHandsome:
I've primarily attempted to go through and make what look to me like monster descriptions more legible and at times humorous. I've only gone down to about ETC_20150317_000771 in this revision.

Some of the monster texts made little sense to me, in which case I was forced to consult a very sub-par translation of the Korean and cross-reference it with the localization I was replacing. If that didn't make sense either, I left it as is to come back to later. A few descriptions have had details trimmed off because I didn't find them important to the line and/or found it too awkward to phrase it with them still there.

In addition, I've adjusted some of the lines near the bottom of the file which appear to be tutorials, tips of the day, and interface text.

I hope I didn't mess up any of the formatting, and I'm not sure what the x prefix on any of the lines means so I didn't remove any. Let me know if I've made a mistake!

### DIFF
--- a/ETC.tsv
+++ b/ETC.tsv
@@ -1,40 +1,40 @@
 ﻿ETC_20150317_000001	Kepa
-ETC_20150317_000002	The Kepa is not a very strong monster. However, attacking it recklessly may bring you to tears.
+ETC_20150317_000002	The Kepa is not very strong, but attacking it recklessly may bring you to tears.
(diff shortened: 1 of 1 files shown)


imcgames:
We will be updating you with details in the near future regarding this matter so please hold your Pull Requests until things clear up. 

PR #664, opened 2015-08-03
Changes i've made so far

AssisrMatheus:
Hi guys, I was playing KR CBT3 then to help the community I started to help getting the translation better since I don't speak korean and can't report bugs.

I just read on the forum that i had to put an X in front of the edited lines, well i already pulled the request so i'll leave these like they're now, butw i'll work on the next ones.

For some reason, the github is extremely slow and of my mistake i wrote Confir and not Confirm, but i won't fix it since it's better to send 50+ lines.

I didn't sent 50+ lines because i won't be able to play as much as i did until now, so i just sent what i've worked.

### DIFF
--- a/ETC.tsv
+++ b/ETC.tsv
@@ -13613,3 +13613,4 @@ ETC_20150717_013612	This game is rated ESRB Teen.
 ETC_20150717_013613	{How} Silvers obtained.
 ETC_20150717_013614	{@st59}Increase Max SP, SP recovery, Block penetration{/}
 ETC_20150717_013615	{@st59}Increase Magic ATK{/}
+ETC_20150729_013624 Battle League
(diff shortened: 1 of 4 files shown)


ETC.tsv
```diff
@@ -2707,7 +2707,7 @@ ETC_20150317_002706	Are you sure you want to cancel the item registration?
 ETC_20150317_002707	Your item has been successfully canceled on the Market. Please check that you've received the item.
 ETC_20150317_002708	The equipped item cannot be registered.
 ETC_20150317_002709	Private
-ETC_20150317_002710	Confirm Purchase
+ETC_20150317_002710	Confir
```
ttgmichael: Confirm?

AssisrMatheus: Exactly, Sometimes github is extremely slow here i don't know why. I just saw it because you saw it. Sorry!

imcgames:
Thank you :+1:

Issue #476, opened 2015-07-13
Marking lines that have been edited


imcgames:
Hi guys!

We've noticed that we are getting multiple pull requests for the same lines a lot recently which doesn't seem like an efficient way for things to go. :scream: 
Therefore, in order to prevent duplicate works on the same line by multiple contributors, 
we thought of marking the lines that have already been worked on for others to just skip those lines and continue on with lines that have not yet been checked. 

We are thinking of using an 'x' in front of the codes that have been edited already,
(so that it will be easy to notice and we can just use 'replace all' before we use the file)

ex: 
ETC_20150317_000001 ->  xETC_20150317_000001
ITEM_20150317_000001 ->  xITEM_20150317_000001
QUEST_20150317_000001 -> xQUEST_20150317_000001

but since this means that we will also ask you to mark the lines as well when you edit one, 
we'd like to know how you think about it before we go on to implement it. 
So please feel free to share your thoughts with us regarding this matter. :smiley: 
(we'd like to implement this ASAP if possible :wink: )

Pomez:
It's a great idea but as a result, some incorrectly edited lines may easily be overlooked. For example I edited some lines but they might not be good enough. Other contributors could have handled them better than me but they just skipped them because "those lines were edited already".

Flevalt:
I've had a similar thought yesterday as well, that if it were a bit more structured, it'd go much faster.

The marks are a very nice idea.
However, simply marking the line after editing it will only mark that the line has already been edited, right? 
It's not an indicator of whether that line is correctly translated, whether it still needs editing or grammar checks.
So basically the mark is more of a priority indicator than an indicator of how complete the line is. 
When you see the mark, you skip those lines in favor of lines that have not been marked yet. 
And this continues until all lines are marked, no matter what state they were left in from their last edit.

I think everyone can agree that the translated/edited lines all differ greatly in matters of quality.
So would it not be even more beneficial to at least use a few different marks?

Something like:

xETC_20150317_000001 = Raw English translation, but still largely incomprehensible. Needs further editing to become comprehensible.
(if you can't understand what the line is trying to say, then this is the state it is in)

yETC_20150317_000001 = Correct idiomatic English translation. Requires further grammar checks
(if you can follow what the line is saying but spelling, punctuation, sentence structure etc. are wrong, then this is the state it is in)

zETC_20150317_000001 = 100% Complete. Minor adjustments needed at most.
[...]

PR #668, opened 2015-08-03
ETC_20150717_013601

MrLiar:
grammar change from "in" to "on" since, "it is already on fire!" makes more sense than "it is already in fire!"


Sourpusss:
Add and X and edit more than 1 line at a time.

MrLiar:
add a X where? and will do on more then 1 line

Sourpusss:
https://github.com/Treeofsavior/EnglishTranslation/issues/476

Add "x"'s to the beginning of edited lines that you deem completely acceptable for the game.  If you're iffy on the line when you're done with it, explain why, and don't put an x.  Also, lines with x's already shouldn't have to be touched - feel free to read them, but unless it's a blatant mistake, don't adjust it.

imcgames:
Seem like it's been fixed in earlier pulls :smile:

PR #669, opened 2015-08-03
QUEST_JOBSTEP.tsv Lines 450-699

Ragunaga:
General grammar, contextual, phrasing, and spelling corrections.

Line 627 is oddly phrased, and out of context. I think it may be wrongly translated. It'd be great if someone could check that, I left it alone.

One line used Mostem instead of fire pot. Changed.

Added 'x's to all the lines I edited, not sure if it's a good idea, but it's in the guidelines.


imcgames:
Thank you :+1:

PR #685, opened 2015-08-04
Update SKILL.tsv

Rezken:
Editing lines 2200-2300:
- Critical grammar mistakes fixed.
- Upper-case fixes.
- Skill name fixes.


imcgames:
:+1:

PR #722, opened 2015-08-06
Update QUEST.tsv (Lines 1-400)

Miyutan:
Grammatical corrections and revising.


Miyutan:
Senhart has already fixed most of these, so I will close this and continue where he left off.

PR #727, opened 2015-08-07
Update QUEST_LV_0100.tsv

ghost:
Checking and editing lines 1854-1914

Edited line 1805 using Requinn's suggestion as well, but it'd still be nice to see it in context.
Line 1913 might need re-translation. The second part in the original Korean line is a question, but the translated line is not.
Also, line 1891's second part says, "How did she unleash the seal?" but I think it might be trying to say something like "How in the world was the seal undone?". Need a second opinion.


imcgames:

PR #679, opened 2015-08-04
Update QUEST.tsv [Lines 1-500]

ghost:
Resending pull request #655 to make sure the changes show.

PR #737, opened 2015-08-07
Update QUEST.tsv [Lines 1-2000]

ghost:
First batch: lines 1-500. Resubmission of pull request #679
The text seems largely the same except the Day of the Divine Tree/Divine Day/Day of Divine Tree is now referred to as Medis Diena
I'm guessing we don't add x's before QUEST anymore?


ghost:
Second batch: lines 500-1000, I mean.

imcgames:
Could you send the pull in smaller batches so that the changes will be visible through github?

Senhart:
How small?

PR #738, opened 2015-08-07
ETC.tsv: changed "Use <class> Attribute" to "Learn <class> Attributes"

Soukyuu:
- "use attributes" didn't make sense, this is a learning NPC
- consistency fix: some classes had "Attribute", some had "Attributes"
- consistency fix: NPC talks about learning attributes, but dialog was saying "Studying attribute..". Changed to "learning attribute"


imcgames:
:+1:

PR #750, opened 2015-08-08
Update UI.tsv [Hair -> Hairstyle]

cordogames:
In a few instances "Hairstyle" makes a lot more sense to me than just "Hair".

Thoughts?


UI.tsv
```diff
@@ -40,10 +40,10 @@ UI_20150317_000039	{@St49}Delete
 UI_20150317_000040	{@St44}Save{/}
 UI_20150317_000041	{@St44}Cancel{/}
 UI_20150317_000042	{@St43}{s22}Hair Shop{/}
```
cordogames, line 42: This could also possibly be changed to "Hairstylist"

ttgmichael, line 42: I prefer Hairstylist :)

imcgames:
:+1:

PR #753, opened 2015-08-08
Update ITEM.tsv [Lines 1-300]

Tkwanshuffle:
Mostly grammatical corrections and a few word changes with a few corrections for better sounding translations.

Line 167 originally states two opposing opinions with regards to the same object. I changed "crafted" to "created", but I think it would be better off with the line reading "How the Hogmas obtained this weapon is unimportant. We are much more concerned with crafting it and using it ourselves." If read like this, then the statement doesn't contradict itself.
 Line 203 needs context. The text sounds weird, but without seeing what context its meant to be in corrections can't be appropriately made.
Line 263 needs context to correct the end (the "type" it references is unclear)

 Opinions?

For humors sake, read line 210 and then 211. "I don't always have a dirty mind...but when I do, I prefer it in text." (dos equis commercial parody in a related context).


imcgames:

PR #756, opened 2015-08-08
ETC changes - Real Time Translation

ttgmichael:
Improving the legibility of many lines, including the lines that seem to be the concatenation of several lines (in Korean its easy, but it caused awkward sentence structures in English).

For example; when you learn a new skill and it's placed in your hotkeys, the message that appears was jumbled, this PR fixes it.

The message that an area's Status is activated is also fixed.

I also renamed "Stats" as "Ability Points" in my game (like Sunaries)

Edits were made from the start of the Game up to the 1st Floor of the Crystal Mines.


imcgames:
:+1: 

(we'll keep it at status points :wink: )

PR #766, opened 2015-08-08
Update QUEST_LV_0200.tsv [Lines 99-550]

asmorgan:
Fixed spelling and grammar of translations, updated some NPC/boss names to be consistent, and adjusted wording to be more natural when it could be done in-context without changing the base meaning.


asmorgan:
Hmm, had in mind to work on editing the translation of this file, but then I saw an earlier pull request for the same file along with commentary on the ellipsis character/3 dots/2 dots issue that I confess myself to have been slightly confused about. Let me know if this editing is untimely and I can close this if necessary.

asmorgan:
And just realized that the changes don't show up. Going to mercy-kill this one, may bring it back in smaller batches later.

PR #772, opened 2015-08-09
Clarified dex stat tooltip stating it affects crit rate rather than crit attack

laurencee:


laurencee:
Forgot to make commit to different branches for the isolated changes

PR #781, opened 2015-08-09
Update SKILL.tsv Problem with Two-Handed being Dual-Wield/Dual-Handed

gillianfrancisco:
Dual-Wield/Dual-Handed changed to Two-Handed (Use Two-Handed Sword/ Two-Handed Sword Mastery)
Lines edited :
4263,4265,4319 skills for two-handed sword confusing for players.
two-handed sword is not dual-wielding swords.


imcgames:
:+1:

PR #786, opened 2015-08-09
Lines 2065-2099

Sourpusss:
2070: The first line was a bit long and redundant of earlier material.
2090: The wording on this is just really weird and complicated.  It needs to be rewritten, and I don't have the time to do that currently.  I left it without an "x."

I replaced all unicode ellipses with just "..." as was agreed upon in the issues.

I change a lot of past tense verbs like "would" and "was" to "will" and "is" (just examples).


imcgames:

PR #804, opened 2015-08-10
ETC.tsv: change Vensi to Banshee

Soukyuu:
Reasoning: 
- drops from Vensi all have Banshee in them, e.g. Banshee Hood, Banshee Matter
- there are other mobs like Red Banshee, etc.

One inconsistency less! No other files in the repo contained "Vensi", so I think that one was the only one.

### DIFF
--- a/ETC.tsv
+++ b/ETC.tsv
@@ -23,7 +23,7 @@ ETC_20150317_000022	Precautions should be taken before fighting this monster.
 ETC_20150317_000023	Green Pokuborn
 ETC_20150317_000024	The meat of this monster is quite nutritious.
 ETC_20150317_000025	Blue Pokubu
-ETC_20150317_000026	Vensi
+ETC_20150317_000026	Banshee
 ETC_20150317_000027	The souls of those who ventured to the underworld now show up as demons.
 ETC_20150317_000028	Red Banshee
 ETC_20150317_000029	Evil Spirit


Soukyuu:
Btw, if someone with more git experience can tell me how to make sure only one commit (the last one) appears in the pull request instead of all of them, feel free to do so :x

What I did was `git pull <official repo> master`, then created a new branch and added the Banshee changes. I thought that would make it list my only commit only, but apparently it isn't so...

ttgmichael:
It's ok, these commits help your git file track the changes made in your repository. Lately the changes made have caused a ton of conflicts as people are prone to making the same line changes with RTT.

Soukyuu:
Yes, but if this one gets merged, won't all those 5 commits appear in the official repository?
I've found [this](http://stackoverflow.com/questions/7244321/how-to-update-github-forked-repository), going to try on my next one.

Nagoda:
One useful trick I have learned (for us who do not speak Korean) is to copy the Hangul of the original word into Google Translate (https://www.google.com/webhp?hl=en#hl=en&q=translate) and use the text-to-voice feature to hear how the word is pronounced. Between the syllables being translated and hearing the pronunciation, you can very often tell when the original word is just a phonetic pronunciation of a word that is actually English.

In this case, the original word is "밴시" and the characters literally are the syllables "Ban" and "Shi", so the name should definitely just be corrected to Banshee.

laurencee:
@Soukyuu Yes you tend to use rebase to replay your commit on top of the existing ones so you don't get a bunch of branch merges.
Either way it doesn't matter too much unless the owner of this Repo says otherwise.

Another problem you have here is you did 2 different commits after each other in the same branch where as those commits should have probably been in different branches since if they accept your request they'll get all those changes from both SKILL.tsv and ETC.tsv in one go...

You should be creating new branches for each individual pull request to prevent this happening.

EDIT: I can see you did create separate branches for your other pull requests but must have forgotten to switch branches before commiting your ETC.tsv changes.

Soukyuu:
I actually switched from master to that branch and then edited the file, I'm not sure why those two other commits appear there. The pull request itself shows 1 file changed, probably because the first one was merged into the official one already, and the 3 following is just me merging stuff from the official one. The last one is the only one intended. I'll rebase the branch like I did with the DEXfix one.

@Nagoda: yes, I actually knew they read it like that. While I don't speak Korean, I know Japanese, and they are similar in some aspects, for example that there is no difference between B/V or R/L. So if you look at it like that, Vensi suddenly becomes Bensi, and knowing some romanisation use si instead of shi, you get to Benshi. From that it's a short way to Banshee.

laurencee:
Looks good now!

imcgames:

PR #805, opened 2015-08-10
ETC.tsv: fix DEX description

Soukyuu:
DEX mistakenly states it increases critical attack but as you can see below, 
it actually increases critical **rate**.

screenshot_20150809_00012

### DIFF
--- a/ETC.tsv
+++ b/ETC.tsv
@@ -4340,7 +4340,7 @@ ETC_20150317_004339	Requires iCoin.{nl}Are you sure you want to reset?
 ETC_20150317_004340	{@st59} Increase in Physical Attack and Critical Attack{/}
 ETC_20150317_004341	{@st59} Increase in HP, HP Recovery rate, Critical Resistance and Max Inventory Slots{/}
 ETC_20150317_004342	{@st59} Increase in Magic Damage and Block Penetration{/}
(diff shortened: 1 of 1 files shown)


Ragunaga:
It increases both I think, dude.

Just not all the time. It increases rate and attack at different rates. Once every other level or so it increases attack a bit.

Soukyuu:
Do you have a screenshot of that? STR states critical attack (= how hard it hits when critical), while DEX seems to be increasing the rate at what it hits.

If it indeed increases both, then the tooltip should state both.

Ragunaga:
Nope, I do not. But I can assure you I've seen it increase both personally.

Maybe someone else can come in and also verify.

Soukyuu:
Well, I'll just leave it to the devs to decide, I guess.

laurencee:
There's a video that proves that DEX affects critical rate and STR affects critical attack so this modification is correct.
https://www.youtube.com/watch?v=L9VEqpXev6M

I've never seen dex affect critical attack. There are lots of attributes and gear in this game that affects crit attack so you may have been doing some other thing like that at the same time.

Regardless, DEX always affects critical rate so the translation that it just affects critical attack is wrong.

imcgames:

PR #808, opened 2015-08-10
Proper translation for SKILL_20150317_002344 (bow mastery related)

merrun:
Change from machine translate to proper translate and no need to put [two handed bow] since bow is already two handed in game, while crossbow are 1 handed


imcgames:

PR #831, opened 2015-08-10
Aiste's story fix.

vtange:
Aiste's story somehow got replaced with the old machine translated text. Here it is again.


ghost:
I've already submitted a rewrite of this story together with my full rewrite of 1001-1138 in my pull request  #743. How do you think we should resolve this?

vtange:
That would work too. Pretty sure your lines 1127 - 1138 were exactly as I left it pre-iCBT. I'm fine with it as long as the lines are no longer machine translated.

PR #855, opened 2015-08-12
weapon name changed to better  

merrun:
the recipe stated it as Ilwoon while the sword itself has different name. 
its an ancient african sword with the same name, don't change the name into something else


imcgames:

PR #822, opened 2015-08-10
QUEST_JOBSTEP.tsv 18, 521, 1478, 1480, 2606, 2905 (Bokor Master gender fix)

Soukyuu:
Bokor Master was referred to as male multiple times. The NPC is female. Also fixed a few sentences I found while searching for the gender ones. Shouldn't interfere with other pull requests.

PR #872, opened 2015-08-20
QUEST_JOBSTEP.tsv 18, 521, 1478, 1480, 2606, 2905 (Bokor Master gender fix)

Soukyuu:
Re-opening #822 after signing the agreement.


Soukyuu:
It seems that all the changes here are already in the bulk upload of QUEST_JOBSTEP, this is now obsolete

PR #891, opened 2015-08-23
UI.tsv [Line 755]

ghost:
Rewrote line to default the with string font parameters


ttgmichael:
Are we sure that the spaces aren't needed? Some lines in ToS actually use spaces in order to center the text.

imcgames:
Yup, @ttgmichael 's comment is the reason why we have some weird spacings in the text :smile:

ghost:
In line 755 of UI.tsv,
The actual setting [Select Language] is using a white font and you barelly can see it.
screenshot_20150824_00003

Are you really sure it was supposed to be white as above?
I know it was supposed to be in the center, but if you use {@st43b}Select Language{/}, the font will be like this:

screenshot_20150824_00002

Issue #900, opened 2015-08-24
New marking scheme and the forums - we're back :)


imcgames:
Hi guys!

We are back from our hiatus, yes, finally, and it's great to be back :smiley: 
As for the issues raised during our time away, I guess we are seeing some progress there too. 
I'd like to take this opportunity to say, once again, we apologize for the ambiguities that were present in using the OTC previously. 
I understand that copyright and license issues were raised out of your concerns for Tree of Savior and thank you for letting us take our time on that before it became too late.

Now, it seems like there are some new guides you've come up with so I'll have that updated in Wiki's guidelines soon. 
And speaking of Wiki, we'll try to have it updated asap too. 
Btw, if there would be changes in Wiki later on, you do not need to do the 'replace all' as we can just do it from here to make stuff easier and have less merge conflicts. 
Maybe I'll let you guys know a day ahead or so before we do 'replace all' so that you can base your edits after we replace stuff when we do them. 

And for the Issue https://github.com/Treeofsavior/EnglishTranslation/issues/807 regarding markings, would it be safe to say that we're good with changing the marking scheme from 'x' in front of the code to '$' in front of the translation? 
If so, we'll update the previous lines with 'x' marks with the new '$' marks. :wink: 

Lastly, there's a 'Translation' category in forums and I want to know what do you guys think of using the forums instead of the 'Issues' here? 
Using Forums or Issues both have its pros and cons such as... Issues can be linked with other issues or PRs but it's really hard to keep up with updates, while the topics in forums are easier to track since topics with new posts are bumped up and lets you know if there are anything new posted but it's more public(?) with the non-translators and I'm not so sure if it can be linked with PR's here :confused: . 
As always, we're open to hearing your suggestions so feel free to share your thoughts about it.

Soukyuu:
I think the issue with markings has been discussed to death already, so we need you guys to just make a final decision. Whichever marking system you decide to go with, as long as we can use the marked files inside the game without having to clean them first, it's all good.

As for communicating on the forums, I think github solves the updates quite nicely. While the issues don't get bumped on new posts, you can just opt to watch the whole repository so you get notifications whenever something happens. Of course, we could use the forums instead, but we'd lose the inter-issue/PR linking to some degree. Or rather, it'd be more work to link, since you have to manually get them instead of just typing #number

Nagoda:
@imcgames 
[...]

Issue #432, opened 2015-07-01
The meaning of a particular term


Sunaries:
While evaluating a pull request of a user, I've come across the term **Divine Day**. 
This is referred from 신수의 날 in Korean. The official site refers this important event as ["The Day of Divine Tree"](https://treeofsavior.com/intro/). 

Maybe @imcgames can elaborate on this? What is the actual meaning of 신수의 날? Divine Day _or_ The Day of Divine Tree? (or should it be The Day of the Divine Tree)?

emailboxu:
https://treeofsavior.com/intro/

I believe it's "The Day of the Divine Tree". "Divine Tree" is a proper noun that appeared in the world and destroyed stuff.

emailboxu:
Errrrrrrrrrrrrrrrm. In that link they refer to it as "The Day of Divine Tree", so maybe it is that? I feel like "The Day of the Divine Tree" makes more sense tbh, but it's their game so..

imcgames:
Yup, @emailboxu is correct. It's the day when the Divine Tree suddenly soared.
'The Day of Divine Tree' in the official site should replace Divine Day. 
(we decided it to be The Day of Divine Tree not so long ago so some other versions of it may exist in the translation files :laughing: we'll be cleaning that up at once later on)

As to The Day of Divine Tree or The Day of the Divine Tree, we chose the former because it is mentioned a lot in the game so we thought we'd go with the shorter one. But if you feel that it is awkward to the point that it must be changed, let us know and we are open to changing it if necessary. :wink:

Nagoda:
"The Day of Divine Tree" does sound awkward I think. "The Day of the Divine Tree" is more natural , but both versions are definitely a mouthful for a term that's used a lot in the game.

"Divine Day" is a little more cryptic but it sounds much better to my ear, and obviously is a lot shorter.

Chaosbelow:
[...]

PR #902, opened 2015-08-24
Update Item.tsv [0~200]

emailboxu:
181 - 신수의 날. https://github.com/Treeofsavior/EnglishTranslation/issues/432

x in front of the lines I've looked over. [#900]

NOTE - I translated these lines with the intention of keeping them as close to the Korean as possible (including some more awkward cases). I did NOT, however, recheck the lines that were grammatically sound and were otherwise fine.


Soukyuu:
The $ should be in front of the _translation_ not the line, as per #900

imcgames:
Thank you :+1: 

(we've changed the markings)

PR #906, opened 2015-08-24
weapon rename to better [asta, eillun, spike ~, morningstar]

merrun:
the recipe stated it as Ilwoon while the sword itself has different name.  (eillun which is incorrect)
its an ancient african sword with the same name, don't change the name into something else

hasta is a type of spear, there's no such thing as asta
morningstar > morning star
spike > spiked

all asta changed to hasta (675, 677, 678, 683)
spike changed to spiked ( 444, 534, 540)
all morningstar changed to morning star (498, 523, 529, 551, 580)


Sunaries:
Morningstar or Morning Star. Both work. It's preference.

imcgames:
Thank you :+1:

PR #910, opened 2015-08-25
Update Item.tsv [201~400]

emailboxu:
Feel like 210 could also be translated as: A rod that benefits anyone studying magic.

252 could also be Altar Rod.


imcgames:
Thank you :+1: 

(we changed the markings :wink:  )

PR #927, opened 2015-08-27
Update SKILL.tsv

Sunaries:
Now, I hope this won't get overwritten again like last time.


grillo-delmal:
Remember to put the $ marks before the translated string (as defined in #900)

imcgames:
Thank you :+1: 

(added $)

PR #932, opened 2015-08-27
Update ITEM.tsv

Sunaries:


grillo-delmal:
Remember to put the $ marks before the translated string (as defined in #900)

imcgames:
Thank you :+1:

PR #934, opened 2015-08-27
ETC.tsv [1-6] [8-24] Proofreading

grillo-delmal:
- Checked blame history and translations to see if someone had submitted one of those lines before.
- Also checked consistency with the machine translator.
- Changed sentences 000004/000006 so that they sound a bit more natural


imcgames:
Thank you :smile:

PR #935, opened 2015-08-27
ETC.tsv [7] Poisonous Kepa

grillo-delmal:
https://forum.treeofsavior.com/t/poisoned-poisonous-or-frenzied/39355/

I opened that thread to discuss the context of that line, most of the people agreed that "Poisonous Kepa" was the option that made most sense. 
We might be missing something since we are translating out of context, if there is any problem with this PR, please join the discussion.

### DIFF
--- a/ETC.tsv
+++ b/ETC.tsv
@@ -4,7 +4,7 @@ ETC_20150317_000003	Large Kepa
 ETC_20150317_000004	Humans sometimes hunt Large Kepa for food.
 ETC_20150317_000005	Red Kepa
 ETC_20150317_000006	Most people are unaware that the Red Kepa is actually quite similar to the Kepa.
-ETC_20150317_000007	Poisoned Kepa.
+ETC_20150317_000007	$Poisonous Kepa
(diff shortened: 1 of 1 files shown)


imcgames:
Well, the kepa is the also the one poisoned/addicted and that's why it had that name :smile:

ttgmichael:
So is a Poisoned Kepa a seperate variant of Kepa or just a Kepa that is poisoned?

If it's a seperate variant, then the change makes some sense. would Toxic Kepa be just as good? Poisonous is a pretty long word.

grillo-delmal:
Ended up reverting it to what it was translated before :) marked it with $ and erased the extra dot that was not in the Korean version.

Staryx:
ttgmichael, toxic kepa would be the same as poisonous kepa but just more poisonous

grillo-delmal:
In the end, IMC responded that the Kepa was sick, not poisonous, so i'll leave it as Poisoned :)

Staryx:
okay my bad :+1:

PR #501, opened 2015-07-17
Update ITEM.tsv

emailboxu:
5485 - 체이퍼러 = Chaferer? Not sure about the Korean here.

5470~

PR #936, opened 2015-08-28
ETC, QUEST, QUEST100: Screenshot Edits

vtange:
Lines
ETC 9348,12545, QUEST 347,501,1000,2031,2279, QUEST100 703,762,1251,1255,1287-88,1562,1722,2277,2869,7649-7655

I made many screenshots during iCBT to both record context and to mark any lines or names I found awkward, before I figured out how to Real TIme Translate. I actually think this way would've worked better since it's less conflict-y than RTT.

We can discuss these edits here or in Discord, where I can put the necessary screenshots easier.


ETC.tsv
```diff
@@ -5380,7 +5380,7 @@ ETC_20150317_005379	Good Job!
 ETC_20150317_005380	Where are you!!!
 ETC_20150317_005381	Come out!
 ETC_20150317_005382	Thorny Gateway
-ETC_20150317_005383	Bag that contains something
+ETC_20150317_005383	$Bag with Something Inside
```
terrapin47: You can actually see the bag, and seeing that it says "Bag with Something Inside" is a bit silly, since you can tell that the bag contains something by just looking at it.  I'd suggest just "Bag".

vtange: The Korean text literally says Bag with something inside, just trying to keep as much meaning as possible.

We can decide if "Bag" works during localization.

grillo-delmal: how about "Bag that contains something"?

Staryx:
quest line 501, the word then, please check if the 2 sentences are in continuation but just in different box, or both of them are 2 different things.

If it is continuation, then would be appropriate i think

vtange:
Quest #501 was a line I worked on pre-CBT. The scene is the very first time your character runs into the speaker, whom is obviously an antagonist of some sort. The word 'then' seems unnecessary for what seems to be a threat from the speaker, trying to question the hero's resolve. That's why I believe the old version works better.

imcgames:
Thank you :smile:

PR #897, opened 2015-08-24
Real Time Translation: Main Quest from Veja Ravine to Kvailas Forest, + Fedimian villagers + Mage Tower

vtange:
Reupload of lines done during iCBT.

PR #896, opened 2015-08-24
ETC, ITEM, QUEST_JOBSTEP, SKILL - "Merog"s and Bramble

vtange:
Standardize name of common mob. Merlog/Merrogg -> Merog.

Special thanks to g1real.

Issue #807, opened 2015-08-10
System for marking lines contributed DURING CBT w/ Real Time Translation.


vtange:
Just like we had "x"'s and stuff for marking lines that have been worked on. I think we will need a system marking lines edited in-game. I'm pretty confident most of these lines will be accurate/usable in game since we did them knowing the context of the game. The worst thing that could happen to this project now is we add translators who had not played the game and they edit lines that were made during the CBT, sending us right back to square one.

Soukyuu:
The problem with the x in front of the line was that the game failed to read the line because of that. We could put an x at the start of the message instead, but tbh that just looks ugly. If the lines were edited with the context of the game already, I think re-editing changing the meaning is not that likely to happen, unless people go back to Korean and re-translate.

Soukyuu:
As mentioned in #790, currently $ is a better candidate, because it's not used by anything in the game. Advantage of my suggestion is that you can
- see the lines in the game without having to edit your files every time
- edit your files from within the game via RTT and see the result immediately (think {nl} placement)
- see which files already have been edited from within the game

imcgames:
Alright we'll check into it and consider changing if there'd be no problem with it :)

Soukyuu:
As pointed out by @Sourpusss, `$` might be a problem if your script uses that to denote variables. I have searched through the current files, and it seems that `§` might be an alternative in that case (unless ToS will contain some sort of law books in the future)

imcgames:
I think we can just use $ for now.
We can place the $ at the beginning of the translation so that you can still see it in game with the $.
We'll then just replace (tab)$ with (tab) for patches in game. 
We're also considering adding the symbol to the line code and making it possible for the translation to be read by client even with the symbol added to the line code. 
However, since we haven't finalized its feasibility and further developments regarding translations, what do you think of adding $ in front of the translation for now?

Soukyuu:
Maybe marking **unedited/untranslated** lines instead would be better. Meaning, seeing a `$` at the start of the translation means the line has not been finalized. This would make editing easier, because once you finished editing, you will see the line as it will appear in the final version. There might be cases otherwise, where linebreaks will shift because of the `$`.

ttgmichael:
@imcgames It's a good idea, that'd really help make sure people know what has been editted and what hasn't.

[...]

PR #938, opened 2015-08-28
Adding $ from previous PRs (#896 & #897)

vtange:
As per https://github.com/Treeofsavior/EnglishTranslation/issues/807

Sorry for any inconvenience. I was going to add them after finding out we settled on a system and then found out my PRs were pulled already.


imcgames:
Thank you :smile:

PR #947, opened 2015-08-29
Update QUEST.tsv [Lines 1651-1800]

ghost:
Lines 1651-1800 reviewed. Lines that disappeared after the merge to CBT re-added. Open to feedback.

Again, changed "contaminating gap" to "corruption breach." If anyone can think of a better keyword or think the original is fine then I'm okay with changing it to that or back to the original.

### DIFF
--- a/QUEST.tsv
+++ b/QUEST.tsv
@@ -1649,151 +1649,154 @@ QUEST_20150323_001648	Huh?
 QUEST_20150323_001649	Huh? 
 QUEST_20150323_001650	Huh? 
 QUEST_20150323_001651	Huh? 
-QUEST_20150323_001652	We are fighting hard. {nl}Don't you think my injured men are more urgent than the farmers? {nl}But this Druid guy provided medicine only to the farmers. {nl}And the farmers themselves also hid the medicine too. I'm speechless…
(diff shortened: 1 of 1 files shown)


QUEST.tsv
```diff
@@ -1649,151 +1649,154 @@ QUEST_20150323_001648	Huh?
...
+QUEST_20150323_001715	$Old Seeds
+QUEST_20150323_001716	$Shaton's Stubbornness
 QUEST_20150401_001717	Star Statue
 QUEST_20150401_001718	Nothing is happening…
-QUEST_20150401_001719	Supressing Altar
-QUEST_20150401_001720	Altar is already activated
-QUEST_20150401_001721	The candlestick is already lit
-QUEST_20150401_001722	Need something to light it…
+QUEST_20150401_001719	$Suppressing Altar
+QUEST_20150401_001720	$The altar is already activated.
```
Sunaries, line 1720: The -> This altar? (since there are many altars in the world of ToS)

ghost, line 1720: I assume the dialogue appears after interacting with the suppressing altar so it should make sense. Not sure.

grillo-delmal, line 1720: If it's a dialogue between 2 characters, "this" would fit, but it seems to be an interaction with the altar, so "The altar" would fit in that context.

ghost:
I guess this is good to go.

imcgames:
Thank you :smile:

PR #948, opened 2015-08-30
Update ETC.tsv

joon-won:
Review from line 1 to line 100

Partly corrected : line 2, line 45. line 46 line 55 line 83 line 85
Retranslated(these needed to be, srsly) : line 4 line 6 line 22 line 24 line 59 line 63 line 65 like 67 line 79 line 96 line 191

-please check line 36 to line 41. Shouldn't the name of the monster be Krumabo or Krumavo? Is there any reason you chose it to be Bushspider?

-for Line 88, I suggest replacing its name(currently Matsum) to be Mesume. The pronunciation of 'Matsum' isn't even close to the pronunciation of the original name. And as well as for line 90, I suggest replacing Matsuma with Mesuma 

-for Line 91, 

-for Line 92, I suggest replacing its name(currently Molich) with Molek. Almost same reason to line 88 and 90 case.

### DIFF
--- a/ETC.tsv
+++ b/ETC.tsv
@@ -1,9 +1,9 @@
 ETC_20150317_000001	Kepa
-ETC_20150317_000002	The Kepa is not a very strong monster. However, attacking it recklessly may bring you to tears.
+ETC_20150317_000002	The Kepa is not a strong monster. However, attacking it recklessly may bring you to tears.
 ETC_20150317_000003	Large Kepa
-ETC_20150317_000004	Humans sometimes hunt Large Kepa for food.
+ETC_20150317_000004	Sometimes there are people who are trying to eat Large Kepa. Don't let them.
 ETC_20150317_000005	Red Kepa
-ETC_20150317_000006	Most people are unaware that the Red Kepa is actually quite similar to the Kepa.
+ETC_20150317_000006 Despite the normal cognition, it is Red Kepa which is more similar to the original Kepa
 ETC_20150317_000007	Poisoned Kepa.
 ETC_20150317_000008	Monster
 ETC_20150317_000009	Hamming
@@ -19,9 +19,9 @@ ETC_20150317_000018	Depending on its maturity this monster may develop a variety
 ETC_20150317_000019	Pokuborn
 ETC_20150317_000020	Yellow Pokuborn
 ETC_20150317_000021	Blue Pokuborn
-ETC_20150317_000022	Precautions should be taken before fighting this monster.
+ETC_20150317_000022	There are many things you should be aware of this monster. Above all, don't get hit by its charge.
 ETC_20150317_000023	Green Pokuborn
-ETC_20150317_000024	The meat of this monster is quite nutritious.
+ETC_20150317_000024	Even after their change into the monsters, their meat is still edible. 
 ETC_20150317_000025	Blue Pokubu
 ETC_20150317_000026	Banshee
 ETC_20150317_000027	The souls of those who ventured to the underworld now show up as demons.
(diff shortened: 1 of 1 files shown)


grillo-delmal:
Some tips for contributing :)
- Check for current "Pull requests" so your changes don't conflict with other pending changes
  - Pull requests can be checked here >> https://github.com/Grillo-del-mal/EnglishTranslation/pulls
  - Your pull request conflicts with https://github.com/Treeofsavior/EnglishTranslation/pull/934
  - You can always comment on other pull requests if you think that the proposed translations are off or they have some mistakes
- When translating a line, put a $ character before the translation (this marking scheme was proposed in #900 so people can know what lines are already translated)
- Before intending to change something like a Monster name, propose it on the Issues section
  - https://github.com/Treeofsavior/EnglishTranslation/issues
- Check the guidelines on the wiki https://github.com/Treeofsavior/EnglishTranslation/wiki/Guidelines
- If you have questions, you can ask in the discord server or in the forum translation section
  - https://discord.gg/0RoYSiES4f4zwAk8 
  - https://forum.treeofsavior.com/c/translations

SHUcream00:
Thank Grillo :)

joon-won:
Thanks Grillo :) I appreciate your help and advice.
Seems it gonna take a while for me to get used to the github system.

ETC.tsv
```diff
@@ -1,9 +1,9 @@
 ETC_20150317_000001	Kepa
-ETC_20150317_000002	The Kepa is not a very strong monster. However, attacking it recklessly may bring you to tears.
+ETC_20150317_000002	The Kepa is not a strong monster. However, attacking it recklessly may bring you to tears.
 ETC_20150317_000003	Large Kepa
-ETC_20150317_000004	Humans sometimes hunt Large Kepa for food.
+ETC_20150317_000004	Sometimes there are people who are trying to eat Large Kepa. Don't let them.
```
ttgmichael: There are people who try to eat the Large Kepa sometimes. Don't let them. ?

ETC.tsv
```diff
@@ -1,9 +1,9 @@
...
 ETC_20150317_000005	Red Kepa
-ETC_20150317_000006	Most people are unaware that the Red Kepa is actually quite similar to the Kepa.
+ETC_20150317_000006	Despite the normal cognition, it is Red Kepa which is more similar to the original Kepa
```
ttgmichael: Contrary to common awareness, the Red Kepa is similar to the original Kepa. ?

ETC.tsv
```diff
@@ -19,9 +19,9 @@ ETC_20150317_000018	Depending on its maturity this monster may develop a variety
 ETC_20150317_000019	Pokuborn
 ETC_20150317_000020	Yellow Pokuborn
 ETC_20150317_000021	Blue Pokuborn
-ETC_20150317_000022	Precautions should be taken before fighting this monster.
+ETC_20150317_000022	There are many things you should be aware of this monster. Above all, don't get hit by its charge attack.
 ETC_20150317_000023	Green Pokuborn
-ETC_20150317_000024	The meat of this monster is quite nutritious.
+ETC_20150317_000024	Even after their change into the monsters, their meat is still edible.
```
ttgmichael: Even after mutating into monsters, their meat is still edible.

ETC.tsv
joon-won: @ttgmichael / In the original text, the writer is comparing "the Red Kepa" to sth he/she didn't mention about. Let me explain the meaning of the original text; (Compared to a certain Kepa like creature), the Red Kepas are "more simillar" to the original Kepas unlike people usually think. Maybe they are talking about their look, maybe they are talking about genes, maybe they are talking about the nature of these creatures, but I don't know. How do you think we should?

ETC.tsv
joon-won: @ttgmichael /  your text better is much better imo, thank you for correcting it

joon-won:
Updated with your suggestions
Thanks again @Grillo-del-mal and @ttgmichael !

grillo-delmal:
Your changes are making conflicts with some of the current accepted changes (specifically lines 2, 4, 6, 22 and 24) You'll need to merge this "pull request" into your branch to get the latest changes.

https://github.com/SHUcream00/EnglishTranslation/pull/1

If you feel that translations of those lines still feel off after merging, you can change 'em afterwards.

grillo-delmal:
And done, good to go ;)

imcgames:
Thank you :smile:

PR #949, opened 2015-08-30
468, tresh to thresh. missed name fix

merrun:
468 tresh to thresh
must have missed it before ...somehow


imcgames:
Thank you :smile:

PR #952, opened 2015-08-31
ETC.tsv [36-41] Bushspider, Bush Spider or Crewmabow

grillo-delmal:
There is a problem with the translation of 크루마보
- On line 36,38,39 & 41 it's translated Bushspider
- On line 40 it's translated as Crewmabow (how 크루마보 is pronounced in English)
- The "Bushspider" don't actually exists, but the "Bush Spider" does... but since this is a fantasy game, this is not too relevant.


imcgames:
Thank you :smile:

PR #954, opened 2015-08-31
Update QUEST_LV_0200.tsv

Sourpusss:
Line 10: "forward unit" was changed to "front line," as I'm pretty sure that's the intended expression.  Forward unit is a made up / improper term, nonetheless.

Instances of Kateen Forest were changed to Katyne, as per ensata's translation guideline (https://github.com/ensata/TreeOfSavior-en-research/blob/master/mapnames.tsv)


imcgames:
Thank you :smile: 

(we will keep Kateen Forest as is for now :wink: )

PR #956, opened 2015-08-31
Update Item.tsv [601~700]

emailboxu:
I went through all of these before but it seems that people went through the same lines and messed with all the translations. I might have to go back to the top and re-edit all the lines that were heavily altered. The lines I'm switching are those that are not direct translations to the Korean, where people have taken great liberties in "translating" (more accurately "localizing") the text. Most likely people read the lines that were direct-translated and altered them without consulting (or understanding) the original Korean. Hopefully such cases will stop in the future.

Line 674 has a mistake in the text, it says "개령형" but should (most likely?) be "개량형".

696 - The original Korean is 2 sentences but one line is much much better in English (fragment otherwise).

Edit; Adding $, almost forgot.


imcgames:
yup, 개령형 seems to be a typo of 개량형

imcgames:
Thank you :smile:

PR #960, opened 2015-09-01
ETC.tsv [25-104] - Proofread

grillo-delmal:
Proofreading monster names, found that "Nest of Gurehachi" was actually a "Grey Beehive" (회색 벌통)


imcgames:
Thank you :smile:

PR #961, opened 2015-09-01
ETC.tsv [105-203] Checking monster names

grillo-delmal:
Checking translations of monster names
- 130: ~~주홍 is Scarlet and was translated as Orange~~ Sunnaries commented that the older translation was correct.
- 140: Changed "Spectre of Dazzled" to "Spectre of Deceit"
- 144: Changed "Dunball" to "Thornball" based on the translation on line 146
- 199: ~~투구벌레 translates to "Beetle"~~ Sunnaries commented that the older translation was correct.
- 203: Since it had the same name as 201 made a small change to show the diference

### DIFF
--- a/ETC.tsv
+++ b/ETC.tsv
@@ -127,7 +127,7 @@ ETC_20150317_000126	$Red Ridimed
 ETC_20150317_000127	$Sakmoli
 ETC_20150317_000128	Attacking its legs will not cause it to topple over.
 ETC_20150317_000129	$Green Sakmoli
-ETC_20150317_000130	$Scarlet Sakmoli
+ETC_20150317_000130	$Orange Sakmoli
 ETC_20150317_000131	$Black Sakmoli
 ETC_20150317_000132	$Rikaus
 ETC_20150317_000133	$Green Apparition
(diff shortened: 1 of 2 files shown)


ETC.tsv
```diff
@@ -102,105 +102,105 @@ ETC_20150317_000101	Big Blue Griba
...
+ETC_20150317_000126	$Red Ridimed
+ETC_20150317_000127	$Sakmoli
 ETC_20150317_000128	Attacking its legs will not cause it to topple over.
-ETC_20150317_000129	Green Sakmoli
-ETC_20150317_000130	Orange Sakmoli
-ETC_20150317_000131	Black Sakmoli
-ETC_20150317_000132	Rikaus
-ETC_20150317_000133	Green Apparition
+ETC_20150317_000129	$Green Sakmoli
+ETC_20150317_000130	$Scarlet Sakmoli
```
Sunaries: It is orange. So it should be reverted back.

ETC.tsv
```diff
@@ -102,105 +102,105 @@ ETC_20150317_000101	$Big Blue Griba
...
-ETC_20150317_000194	Boowook
+ETC_20150317_000194	$Boowook
 ETC_20150317_000195	One of the defensive systems left in the king's tomb.
-ETC_20150317_000196	Bearkaras
-ETC_20150317_000197	Pino
+ETC_20150317_000196	$Bearkaras
+ETC_20150317_000197	$Pino
 ETC_20150317_000198	Early pioneers came up with its name from its similarity to the Gepetto.
-ETC_20150317_000199	Helmet Bug
+ETC_20150317_000199	$Beetle
```
Sunaries: The previous name 'Helmet Bug' was correct. It has 'helmet' (투구) in it and 'bug' (벌레) in it.

imcgames:
Thank you :smile:

PR #975, opened 2015-09-05
Update ETC.tsv

Sunaries:
Minor fixes.


ETC.tsv
```diff
@@ -1477,8 +1477,8 @@ ETC_20150317_001476	Team does not exist
 ETC_20150317_001477	{S18}{ol}{#ffffcc}This character does not exist
 ETC_20150317_001478	Cannot start game
 ETC_20150317_001479	{S18}{ol}{#ffffcc} Disconnected from the server due to verification failure
-ETC_20150317_001480	You have exceeded the maximum weight and cannot receive item. {nl}Sell or discard items in inventory to reduce weight.
-ETC_20150317_001481	You cannot receive the item. You have exceeded the maximum number of items allowed
+ETC_20150317_001480	You cannot obtain items anymore, because you have exceeded the maximum weight you can carry. {nl}Sell or discard items from your inventory to reduce weight.
```
laurencee: The comma before "because" is unnecessary:
"You cannot obtain items anymore because you have exceeded the maximum weight you can carry."

Sunaries:
:+1: Thanks. Good to merge.

ETC.tsv
```diff
@@ -1477,8 +1477,8 @@ ETC_20150317_001476	Team does not exist
...
+ETC_20150317_001481	You cannot obtain items anymore. You have exceeded the maximum number of items allowed.
```
Soukyuu: Should use the same wording for this and 1480 to make it look more uniform - only the reason changes. For example: "You cannot obtain any more items. You have exceeded the maximum number of items allowed."

ETC.tsv
Soukyuu: Alternatively: "You cannot obtain any more items. You have exceeded the maximum weight you can carry."

Sunaries:
Good to merge.

I've changed **ETC_20150317_001480** to just let the player know that they've exceeded their maximum carrying weight. Adding more text to the first sentence would only create a new line which just looks ugly on the screen.

imcgames:
:+1:

PR #980, opened 2015-09-05
1-249 ETC.tsv (Corrected lots of $'d lines - be more careful, please)

Sourpusss:
Went through the first 250 lines, attempting to finish any missing $'d lines, only to find this section was still riddled with errors.  I touched up a handful of stuff while I was at it, but a decent portion of this is error fixes for lines already marked with $'s.  A lot of stuff like this shouldn't pass through review - it's wasting all of our time if it does.

---

Line 124: "an herb" is proper for American English.

Line 158: changed "ruins" to "tombs," as it makes more sense in the context of the ambiguity as to when they were placed.
     -Based on the description of the Boowook, it seems like these "defensive creations" were created when Zachariel the Great was still
      around.  That being said, the Royal Mausoleum wouldn't have been ruins at the time of their creation (right?).

Line 162: same as above.

Line 191: I honestly view the plural of Vekarabe to be best kept identical to its singular form.  A knitpicky choice of mine, but correct me if I'm wrong here.

Line 195: Changed "is" to "are," for a similar reason to the above change.  Please correct me if this is not right / you don't agree with this fictional grammar.

Line 200: The description seemed unfinished? I'm not sure, but I went ahead and added a second part the sentence that made it seem way more complete.  Also used an idiom.

Line 216 / 217: Changed "Danderu" into "Dandel" for uniformity.  Please correct me if that was the wrong term to change, but more items and their descriptions are already aligned to the term Dandel.

### DIFF
--- a/ETC.tsv
+++ b/ETC.tsv
@@ -3,9 +3,9 @@ ETC_20150317_000002	$The Kepa is not a strong monster. However, attacking it rec
 ETC_20150317_000003	$Large Kepa
 ETC_20150317_000004	$There are people who try to eat the Large Kepa sometimes. Don't let them.
 ETC_20150317_000005	$Red Kepa
-ETC_20150317_000006	$Contrary to common awareness, it is Red Kepa which is more similar to the original Kepa
-ETC_20150317_000007	$Poisoned Kepa.
-ETC_20150317_000008	$Monster
+ETC_20150317_000006	$Contrary to common belief, it's Red Kepa that are more similar to the original Kepa.
+ETC_20150317_000007	$Poisoned Kepa
+ETC_20150317_000008	Monster
 ETC_20150317_000009	$Hamming
 ETC_20150317_000010	$Despite its tough skin, the Hamming is a fairly feeble monster.
 ETC_20150317_000011	$Chinency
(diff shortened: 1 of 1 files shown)


ETC.tsv
```diff
@@ -3,9 +3,9 @@ ETC_20150317_000002	$The Kepa is not a strong monster. However, attacking it rec
 ETC_20150317_000003	$Large Kepa
 ETC_20150317_000004	$There are people who try to eat the Large Kepa sometimes. Don't let them.
 ETC_20150317_000005	$Red Kepa
-ETC_20150317_000006	$Contrary to common awareness, it is Red Kepa which is more similar to the original Kepa
-ETC_20150317_000007	$Poisoned Kepa.
-ETC_20150317_000008	$Monster
+ETC_20150317_000006	$Contrary to common belief, it's Red Kepa that are more similar to the original Kepa.
```
Soukyuu: I personally don't see a reason to go from "it is" to "it's".
Other descriptions speak of the monsters in singular, not plural. Let's keep the number consistent.

ETC.tsv
```diff
@@ -56,7 +56,7 @@ ETC_20150317_000055	$Some people say a Meduja can be used to contact the souls o
 ETC_20150317_000056	$Blue Meduja
 ETC_20150317_000057	$Green Meduja
 ETC_20150317_000058	$Red Meduja
-ETC_20150317_000059	$Researchers asserts that the plants have souls. And, the Medujas are the mutated form of their souls.
+ETC_20150317_000059	$Researchers have asserted that plants have souls. The Medujas are the mutated form of such souls.
```
Soukyuu: Change to singular?

Soukyuu:
There are a couple more cases of singular vs plural when describing monsters, I think we should decide on one way of talking about monsters in descriptions and stick with it.

ETC.tsv
Sourpusss: Not sure what you mean by the singular vs plural in this context? And I occasionally just like to add contractions to keep things from getting stale (by default, there are none).

would "it's the Red Kepa that are" be better?  Also, I agree on Meduja, and I'll change that when I change this.

Soukyuu: Well, I mean, when you write "it's the Red Kepa that are" (= plural), vs the old version that should have been changed to "it's the Red Kepa that **is**" (= singular). I don't mind either way, but as I said, we probably should decide on a consistent way to talk about them.

Sourpusss: Oh oh, I see.  In this case, I feel like referring to them as Kepas is definitely a better way for indicating them being plural, and I'm pretty sure the usage of "Kepas" is much more common in various parts of the translation already, so I'll just put it back to that for simplicity's sake.  But yeah, we need to figure it out.

imcgames:
Thank you :smile: 

(Line 158 and 162 : 
The Korean text mentions it as ruins because the description seems to be based on the present point of view.
I think it is also because we can't rule out the possibility that they'd be placed in some other ruins in the future, which are not necessarily of tombs, so it may have been mentioned as ruins instead of tombs for convenience sake. 
For now, tomb or ruins doesn't seem like a big issue so we'll have it merged  :wink: )

PR #987, opened 2015-09-07
Update ITEM.tsv

Sunaries:
- Removed the extra space for 'Costume', 'Recipe' and 'Card' in the inventory UI.

EDIT: @imcgames This entire file 'ITEM' consists of 1 extra space after each word and sentence. Is this supposed to be? Because for 'Costume', 'Recipe' and 'Card' it isn't. I haven't done everything, only a small portion, because it requires some time to remove all the extra spaces since there are so many lines.

### DIFF
--- a/ITEM.tsv
+++ b/ITEM.tsv
@@ -2719,107 +2719,107 @@ ITEM_20150317_002718	Synthetic Stone
 ITEM_20150317_002719	A material used for item synthesis. 
 ITEM_20150317_002720	String Amulet 
 ITEM_20150317_002721	Magic Amulet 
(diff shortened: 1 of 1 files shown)


Sunaries:
Good to merge, by the way.

imcgames:
Yup, 1 space at the end is intended so it shouldn't be removed

laurencee:
I noticed the last commit actually has useful changes. Probably quicker to just to do a new PR with just those useful changes.

Soukyuu:
@imcgames: so instead of fixing **one** space in your message code, you put **thousands** of spaces into item.tsv. That's... doing it really backwards.

PR #990, opened 2015-09-08
Update ITEM.tsv

Sunaries:
- Removed the unnecessary extra space for 'Costume', 'Card' and 'Recipe' in the inventory UI for consistency. 

GitHub Logo

### DIFF
--- a/ITEM.tsv
+++ b/ITEM.tsv
@@ -1454,7 +1454,7 @@ ITEM_20150317_001453	Skill 2 Bracelet
 ITEM_20150317_001454	Skill 3 Bracelet 
 ITEM_20150317_001455	Skill 4 Bracelet 
 ITEM_20150317_001456	Highlander Costume 
-ITEM_20150317_001457	Costume 
+ITEM_20150317_001457	Costume
 ITEM_20150317_001458	It was made with elegance for the royal ball and formal events while not being restrictive to the movements of your limbs. 
 ITEM_20150317_001459	Peltasta Costume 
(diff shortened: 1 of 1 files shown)


Sunaries:
Good to merge. **However**, @imcgames you told me in https://github.com/Treeofsavior/EnglishTranslation/pull/987 that the extra space in all the lines of ITEM is intended. How do you explain this? 

GitHub Logo

↑ **'Cockatrice Fat 'It's not enough!**

Because of the coding, the notice message is also wrong in English. Because it links the item name (코카트리스의 비계) and the warning (가 충분하지 않습니다) together. In English, this is really difficult to pull of. So, for now, I have changed it to 'is not sufficient'. -> Example: _'Cockatrice Fat' is not sufficient!_

Another example when using Gem Extraction:

Github Logo

Now, I don't think the extra space is intended, because as you can see from the screenshot, you can clearly see that it doesn't look the way it should look like.

imcgames:
Unlike quest lines, item names are usually used more than once, in different ways.
The images above show cases of when item names can be shown properly even without the space and it'd be really nice if all instances of item names used done this way but unfortunately, there are many cases where without the extra space, the output in client would be weird or awkward. 
And because having one extra space seemed better than having cases with no space (yeah..  we had to compare and choose :scream: ) we decided to go with having an extra space.

imcgames:
Thank you :smile: 

(Merged these three lines for now as they seem to be used as labels, not as actual item names. :wink: )

PR #994, opened 2015-09-08
QUEST #1203-1217

vtange:
One last chunk to go for this story before I move on to translate the other leftover lines in QUEST (and finish it once and for all), though you can pretty much guess out what happens next after reading this.


Soukyuu:
Looks good to merge to me.

imcgames:
Thank you :smile:

PR #1000, opened 2015-09-11
Funny Error Fix and Setting Universal Mob Names

Sourpusss:
2418/2419: Changed Fondel to Pawndel for monster naming consistency, and preventing a very humorous error.

2429: So, I know the wiki refers to this boss as Spector (a typo of Specter).  The in-game name for the monster is Specter Master, and some quest lines referring to it this was.  However, a handful of the translation's lines mention it as "Spector monarch" or "Specter Monarch."  I changed it to Specter Master here, but if I could get a definitive answer on an agreed upon name, I can go ahead and spot correct all its references to be universal.  Regardless of what's chosen (Specter, Specter Master, or Specter Monarch) it shouldn't be called Spector, as this is a spelling error as well as one of the (currently) least referenced of names for this boss.

### DIFF
--- a/ITEM.tsv
+++ b/ITEM.tsv
@@ -2415,18 +2415,18 @@ ITEM_20150317_002414	Rodelin Cloth Piece
 ITEM_20150317_002415	Fonde Hair 
 ITEM_20150317_002416	Fonde Bow 
 ITEM_20150317_002417	Fondel Hair 
-ITEM_20150317_002418	Fondel Shaft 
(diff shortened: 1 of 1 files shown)


imcgames:
I've changed the wiki for Specter. Thanks for pointing that out :+1: 
Specter Master or Specter Monarch both works but I guess monarch would be better so as not to confuse it with class masters?

Sourpusss:
Yeah, I think Monarch works best too, but I just put it to Master since it was like that already.  But the issue of class masters is good.  I'll hopefully get that uniform by the end of the day!

imcgames:
Thank you :smile: 

(changed master to monarch :wink: )

PR #890, opened 2015-08-23
Update QUEST.tsv [Lines 1-500]

ghost:
Major rewrite.
I've done this until line 2275 which I'll submit in batches through other pull requests in the next few days after I do some final checks.

Open to feedback.

PR #1009, opened 2015-09-12
Update QUEST.tsv [Various Lines from 1-300]

ghost:
Just a few tweaks that I couldn't make before #890 was pulled. Reverted Line 90 to a previous translation that I mistook for a very liberal translation but seems to be the actual line.

### DIFF
--- a/QUEST.tsv
+++ b/QUEST.tsv
@@ -57,7 +57,7 @@ QUEST_20150317_000056	By royal command, we've come to this forest to investigate
 QUEST_20150317_000057	In the past, pilgrims travelled through this forest. {nl}But now, only soldiers gather here due to the monster presence.
 QUEST_20150317_000058	Up until four years ago, the Goddesses often blessed us with signs of their presence. {nl}But now, even our combined prayers don't bring so much as a tremor. {nl}Have they truly abandoned us?
 QUEST_20150317_000059	Vacenin personally goes to each region to deal with things. {nl}He gets uncomfortable when he stays in the city.
-QUEST_20150317_000060	Do you plan on staying or enlisting? {nl}If you're interested, I can see about moving you up.
+QUEST_20150317_000060	Do you plan on staying or enlisting? {nl}If you're interested,  I'll gladly recommend you to the head commanders.
 QUEST_20150317_000061	Don't let your guard down. Monsters can appear any time, anywhere.
 QUEST_20150317_000062	Where in the world could the Goddess Zemyna have gone…
 QUEST_20150317_000063	What in the world do you call this?
--- a/QUEST.tsv
+++ b/QUEST.tsv
@@ -15,13 +15,13 @@ QUEST_20150317_000014	Goddess Laima
 QUEST_20150317_000015	By the time you get this message, much time will have passed. {nl}I am the Goddess of Fate and Wisdom, Laima.
 QUEST_20150317_000016	600 years ago, I foresaw a lot of grief and death. {nl}If you're seeing this message, then I'm afraid all I've foreseen has likely come to pass.
 QUEST_20150317_000017	The demons have started the first of the cataclysms, and all the Goddesses have vanished. {nl}I knew this would happen and did all I could to change this fate, but my power alone was not enough.
-QUEST_20150317_000018	Instead, I wandered through many far-seeing dreams for an answer. {nl}You were at the end of them… the Savior.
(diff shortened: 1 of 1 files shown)


QUEST.tsv
```diff
@@ -15,13 +15,13 @@ QUEST_20150317_000014	Goddess Laima
...
 QUEST_20150317_000019	However, knowing of my clairvoyance, the demons would never rest until they saw me gone. {nl}I was forced to hide away, until the time of your appearance.
 QUEST_20150317_000020	I resolved to divide my being and subsist inside these many messages. {nl}I continued to secretly create them, even as the kingdoms flew by over the millennia.
 QUEST_20150317_000021	Savior, these messages are scattered across axes of time. {nl}Once they're all brought together, I will regain my complete form and help against the demons.
 QUEST_20150317_000022	This divine task may be a heavy burden, but it is something you must bear. {nl}Perhaps, in the muddy midst of your trials and tribulations, something may come about.
 QUEST_20150317_000023	Nevertheless, only you and I can save this world.
-QUEST_20150317_000024	I have passed the next revelation to the first Paladin. {nl}I'll be waiting for you at the highest flower garden.
+QUEST_20150317_000024	I've passed the next revelation to the first Paladin. {nl}I'll be waiting for you at the highest flower garden.
```
Soukyuu, line 24: I don't see why this change was necessary.

QUEST.tsv
```diff
@@ -56,14 +56,14 @@ QUEST_20150317_000055	I don't know where to start. {nl}Have you heard anything?
 QUEST_20150317_000056	By royal command, we've come to this forest to investigate some strange events. {nl}Witnesses have reported something with a big face and bony fingers. {nl}What in the world could that be?
 QUEST_20150317_000057	In the past, pilgrims travelled through this forest. {nl}But now, only soldiers gather here due to the monster presence.
 QUEST_20150317_000058	Up until four years ago, the Goddesses often blessed us with signs of their presence. {nl}But now, even our combined prayers don't bring so much as a tremor. {nl}Have they truly abandoned us?
-QUEST_20150317_000059	Vacenin personally goes to each region to deal with things. {nl}He gets uncomfortable when he stays in one place for too long.
-QUEST_20150317_000060	Do you plan on staying or enlisting? {nl}If you're interested, I'll gladly recommend you to the head commanders.
+QUEST_20150317_000059	Vacenin personally goes to each region to deal with things. {nl}He gets uncomfortable when he stays in the city.
+QUEST_20150317_000060	Do you plan on staying or enlisting? {nl}If you're interested, I can see about moving you up.
```
Soukyuu: The new version sounds as if they are suggesting to move you up the ranks, instead of recommending your to the ones in charge (as a recruit).

QUEST.tsv
```diff
@@ -87,13 +87,13 @@ QUEST_20150317_000086	The Goddess Vakarine left us shortly before Medis Diena. {
 QUEST_20150317_000087	{memo X}That's right. The strangest one of all was the Goddess Laima. {nl}No one's really sure when she disappeared.
 QUEST_20150317_000088	{memo X}Besides, isn't Laima the Goddess of Fate and Wisdom? {nl}She probably knew all about it.
 QUEST_20150317_000089	Resting Soldier
-QUEST_20150317_000090	I was an adventurer myself just like you, but then I got hit by an arrow…
+QUEST_20150317_000090	I used to be an adventurer like you. Then I took an arrow in the knee…
```
Soukyuu, line 90: I'm guessing that's the "liberal" translation... I wonder if they're allowed to make references to other games just like that.

QUEST.tsv
ghost: The original is the version of the first translation and the one I changed it to is the translation done by ensata. I'm not quite sure which one is more accurate.

Soukyuu: What I'm getting from the KR->JP translation is closer to the first translation, but different:
"Could it be that you are thinking about not staying here and becoming a recruit? If you have such thoughts, I could talk to the upper ranks"

I've asked @SHUcream00 about it on discord.

QUEST.tsv
Soukyuu, line 90: According to SHUcream, it's really the Skyrim meme.

QUEST.tsv
Soukyuu: The original translation is correct, the new one seems like a mistranslation, confirmed by SHUcream on discord. Please revert.

ghost:
Good to merge. :+1:

imcgames:
Thank you :smile:

PR #1001, opened 2015-09-11
QUEST #1218-1235

vtange:
End of Modestas story. Rough translation, editing can come later.
Will be working on the remaining lines mentioned in Issue #922 in another branch

PR #1011, opened 2015-09-14
Double Quotes -> Single Quotes (prevent ingame display error)

vtange:
According to what @Soukyuu said in #1001 , double quotes are a no-no, so I went back and replaced all the quotes I put in during Aiste's story and Modestas's story. There was one extra line (#1959) that I found double quotes in.


imcgames:
Thank you :smile:

PR #1015, opened 2015-09-15
Update SKILL.tsv

Sunaries:
Minor fix to SKILL.
- Changed a few skill descriptions.

@imcgames: http://imgur.com/a/Ehc0o - Here is an album containing a few screenshots of skills where you can see cut-off variables in the tooltip.

GitHub Logo 

You can also see that long attribute names like the one above will overlap the UI. I can't make the name shorter since this is already at its best. Though, for this particular attribute name, I could remove 'basic attack' since it already tells you the effects in the description below. 

**A question**: Is there a way for me to try out different fonts in the game? I know that the game uses 2 fonts at this moment. imcm_book & imcm_original. 

I just would like to see what the game looks like with the following fonts:
- Raavi
- PF Din Text Cond Pro
- JejuGothic
- Noto Sans

Or maybe you can give me a preview with those fonts, please. :stuck_out_tongue_winking_eye:

### DIFF
--- a/SKILL.tsv
+++ b/SKILL.tsv
@@ -4206,9 +4206,9 @@ SKILL_20150714_004205	$Attack: #{SkillAtkAdd}#{nl}Attribute Attack: #{CaptionRat
 SKILL_20150714_004206	$Attack: #{SkillAtkAdd}#{nl}Attribute Attack: #{CaptionRatio}#%{nl}Can use while riding
 SKILL_20150714_004207	Fly circling around target area. Range defense of enemies within target area decreases.
 SKILL_20150714_004208	$AoE Defense Ratio: -#{CaptionRatio}# {nl}Duration: 10 seconds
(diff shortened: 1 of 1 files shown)


Soukyuu:
Ideally, when a string is too long, it should autowrap, not just run out of the tooltip. I think there are already tags that control alignment (left, center, right), so in this case it would autowrap nicely and increase the height of the tooltip by a bit.

imcgames:
Replace  imcm_book & imcm_original with the fonts you'd like to use :wink:
( you need to rename the font file to imcm_book & imcm_original )

Sunaries:
@imcgames Thank you. It worked. Is the font used in the 1st iCBT final? Or can we still send in suggestions with screenshots of course? 

Good to merge by the way.

imcgames:
@Sunaries  I meant the one in release/languageData/English/font :smile:

imcgames:
Thank you :smile:

PR #1016, opened 2015-09-15
Spector -> Specter // Master -> Monarch

Sourpusss:
Going around doing some spot fixes for the naming of this boss (and general spelling error correction for any reference of this type of mob), which was agreed upon to be "Specter Monarch," after a recent merge.

Lines 4514, 4515, 4516, and 4518: I didn't change since it looked a bit like a scripting issue, so I didn't want to touch it.


imcgames:
Thank you :smile:

PR #1019, opened 2015-09-15
Specter Monarch Fixes (Change the Wiki page to include Monarch, too?)

Sourpusss:
I mean, I know it directly translates to just Specter, but for the sake of preventing all of this from getting reverted, maybe add the title to the wiki name?


imcgames:
Thank you :smile: 

(I've added Specter Monarch to wiki :wink: )

PR #1026, opened 2015-09-16
QUEST #2332-2384, Modestas Story finish and last two empty lines

vtange:
Definitely should be leaving {memo X} lines alone now, they are certainly replaced if not by the same line, then by a line with different wording/length. 

With this, I hope QUEST.tsv is just done. Will need to go through it once more for {nl} checks if we can get an in-game demo or something, but until then I'm pretty sure this file can be marked off the list.


imcgames:
Thank you :smile:

PR #1033, opened 2015-09-18
Update SKILL.tsv

Sunaries:
Minor fixes.

@imcgames 

I am not sure if this is intended. But some status effects/skills from monsters are affected by the imcm_book font. So, if you were using a handwritten font appropriate for books, then the status effect/skills from monsters would show the same font.

Below is an example showing this. On the left side, you see a handwritten font and on the right side you see a different font (Cabin Medium) used as imcm_book. 

GitHub Logo

If you are interested in the Cabin Medium font, you can check the screenshots I took with it by [clicking here.](http://imgur.com/a/8jFMX) It actually solves the line space issue that we have with the default font.


Sunaries:
Good to merge.

imcgames:
Thank you :smile: 

(Thanks for sharing with us the imcm_book font issue and the cabin font, I will check on it with the devs.
But in any case, even if we don't get to change the fonts, players may use their own font desired :wink: )

PR #1046, opened 2015-09-22
Update ITEM.tsv [717-737]

Anasu:
I didn't want to include the lines 727 and 728 because I'm not sure about those translations.


imcgames:
Thank you :smile:

PR #1085, opened 2015-09-28
Rukas Plateau -> Plateau of Rex

vtange:
Hoping no conflict. Mausoleum 1F next.

EDIT: Whoops, got the names mixed up. this is for Rukas


QUEST_LV_0100.tsv
```diff
@@ -5826,7 +5826,7 @@ QUEST_LV_0100_20150317_005825	Check the stone pillars of Alkune Uphill
 QUEST_LV_0100_20150317_005826	Now that you have taken care of the monsters, check the stones pillars around Alkune Uphill. You will be able to find traces of relics in the pillars.
 QUEST_LV_0100_20150317_005827	Retrieved all the relics of Raudi. Go back and give the relics to Raune.
 QUEST_LV_0100_20150317_005828	Check the Epitaph
-QUEST_LV_0100_20150317_005829	There are epitaphs about the mausoleum in Shirnas Plateau. Check the epitaph of Shirnas which is one of it. 
+QUEST_LV_0100_20150317_005829	There is an epitaph about the mausoleum in Shirnas Highland. Check the epitaph found there.
```
Sunaries, line 5829: mausoleum -> Royal Mausoleum
epitaph -> Maybe an idea to change this to "inscription"? They're synonymous, but I think "inscription" is a more common and known term to users?

QUEST_LV_0100.tsv
```diff
@@ -8143,6 +8143,6 @@ QUEST_LV_0100_20150918_008142
 QUEST_LV_0100_20150918_008143	
 QUEST_LV_0100_20150918_008144	
 QUEST_LV_0100_20150918_008145	
-QUEST_LV_0100_20150918_008146	
+QUEST_LV_0100_20150918_008146	$Find Historian Rexipher.
```
Sunaries: Remove the period.

QUEST_LV_0100.tsv
vtange, line 5829: I would leave it as epitaph to preserve detail. Inscription is a much more broad word and we're clearly reading from tombstones. ETC also has the map objects labeled as epitaphs. Plus we don't know but maybe our audience of mostly 20-something-yr olds does know that word. :)

vtange:
Should be good to merge.

imcgames:
Thank you :smile:

PR #1088, opened 2015-09-29
realtime update SKILL

terrapin47:
The lines for the Diev skills are also used on map locations, so that a map location might read "Carve Goddess Zemyna".  Reverts edit adding "Carve" to the line so this doesn't happen.


Sunaries:
Good to merge.

imcgames:
Thank you :smile:

PR #1096, opened 2015-09-30
Update ETC.tsv (added latest KOR lines)

Sunaries:
Added the latest lines from the _kor file. The actual _kor file and the rest has to be updated by IMC.

@imcgames 

I noticed a new line that says: {Level} 레벨 - Where does this line apply to? It didn't seem to work on the skill and item descriptions. It would be great to have it for the skill and item descriptions, so that we can change it to "Level {Level}".


terrapin47:
:+1:

Sunaries:
Good to merge.

Sunaries:
Good to merge.

imcgames:
Thank you :smile:  

(the latest lines added in Nexon client aren't necessarily the final version of lines that is used by English version so we removed the added lines :wink: )

PR #1097, opened 2015-09-30
Update UI.tsv (added latest KOR lines)

Sunaries:
Added the latest lines from the _kor file. The actual _kor file and the rest has to be updated by IMC.

@imcgames 

GitHub Logo

Since the 9/30 patch, the text on this button has changed to 정리. Even though it's still "Sort" in the English file, it doesn't seem to be affected by it. Did the developer change it into an image? Error?


terrapin47:
:+1:

Sunaries:
Good to merge.

imcgames:
we wont be using these lines now but we'll keep it and use it later on :wink:

PR #1101, opened 2015-09-30
Updated language files based on current client files

TOSBase:
I updated the korean and english language files based on the current client files. There was some inconsistency in the use of {memo X} between the korean and english files.


terrapin47:
Looks good to me :+1:

imcgames:
Our reference for lines will be ones in the Github repository, not necessarily the ones found in KR client.
(because the lines in KR client may be different sometimes so we won't be merging this today :wink: )

PR #1116, opened 2015-10-01
Update QUEST.tsv

brencoy:
Fixed grammar errors relating to the use of the word 'already'.

-The word 'already' cannot be used to end a sentence, unless it comes before the verb 'be'.
-The word 'already' has to used before a noun or verb.


vtange:
If you're new here, please refrain from changing lines with "$" and {memo X}. The former are lines that have been hand-worked by people WITHIN the game itself and the latter are lines no longer used in-game. IMO we don't need to pay too much attention to exact word use as long as it gets the point through clearly.

Also, I'm pretty sure 'already' would have absolutely no way of being at the end of a sentence if it has to follow the first rule you wrote. http://blog.oxforddictionaries.com/2011/11/grammar-myths-prepositions/

brencoy:
Thank you. I apologize, "$" and {memo X} being unused is not clarified in the translation wiki, so I had no idea.

PR #1153, opened 2015-10-04
Contaminated, Polluted -> Corrupted Standard

ttgmichael:
This PR changes most contaminated and polluted instances into corrupted.

I didn't like how the word "Contaminated" was used in a lot of places. For one, it's a very long word, and another reason is that the word more or less connotates hygiene or cleanliness. And many things in TOS aren't being sullied by dirt or oils, but rather demonic energy.

So I want to suggest changing the term to corrupted or tainted. I do realize there is a boss named "the Corrupted", so I'm open to suggestions on what word to use rather than contaminate.


terrapin47:
Good idea, I was wondering about the instances of "contaminated" things myself.  
Changes look good :+1:

ETC.tsv
```diff
@@ -4789,8 +4789,8 @@ ETC_20150317_004788	{memo X}I found the information about the flow of horsepower
 ETC_20150317_004789	{memo X}Special information cannot be found.
 ETC_20150317_004790	Please remove the trap which attracts the monster!
 ETC_20150317_004791	{memo X}More monsters are approaching, it is dangerous.
-ETC_20150317_004792	Purifying contaminated area
-ETC_20150317_004793	Purified demon's contaminated area!
+ETC_20150317_004792	Purifying the Demon Corrupted Land
```
Sunaries: Purifying corrupted land ?

ETC.tsv
```diff
@@ -4789,8 +4789,8 @@ ETC_20150317_004788	{memo X}I found the information about the flow of horsepower
...
+ETC_20150317_004793	Purified the Demon Corrupted Land!
```
Sunaries: Purified the demon corrupted land! ?

ETC.tsv
```diff
@@ -916,7 +916,7 @@ ETC_20150317_000915	Woodfung
 ETC_20150317_000916	The Woodfung is closely related to witchcraft.
 ETC_20150317_000917	{memo X}Red Infrorocktor
 ETC_20150317_000918	They exhibit a unique hull that is a mix between stone and iron.
-ETC_20150317_000919	Contaminated Vikaras Mage
+ETC_20150317_000919	Corrupt Vikaras Mage
```
Sunaries: Not "corrupted" ?

ttgmichael: Corrupt can be used as an adjective. To say corrupted would just make it use its verb form (to become corrupt).

terrapin47: I think he's right though.  Aren't these the passed out mages in the mausoleum?  They've been "Corrupted", and are in that state, rather than mages that were once good, but are now Corrupt and fighting against you?

ttgmichael: oh ok. I'll change it to match the context ( never fought them before so I didn't know >< )
Thanks for the info!

vtange:
If you get tired of that word, you can also used 'sullied'.

Been doing this for my own PRs since Sirdgela Forest in iCBT1 (contaminated altar -> corrupted altar)

Good work, hope there's no conflicts

Sunaries:
Good to merge.

imcgames:
Good idea. Thank you :+1:

PR #1172, opened 2015-10-06
Update ITEM.tsv

Azioria:
Translations for lines 7030-7032, 7055-7058, 7062, 7064-7068, 7071-7082 (25 lines)


imcgames:
"결계" is mean like "barrier" so it should be changed to like "barrier stone". What do you think?

Azioria:
I think that sounds more accurate, as well. I've switched that as well as fixed lines 7026-7027 to be more consistent.

Sunaries:
Looks good after the change. :+1: Good to merge.

imcgames:
Thank you :+1:

terrapin47:
Hi @Azioria, would you like to join us on Discord?  We're using it to discuss changes and help each other with the translation process.  You can join by clicking [here](https://discord.gg/0RoYSiES4f1LlZPF).

PR #1178, opened 2015-10-06
ETC - Corrupted Standard, Transport Magic Circle

vtange:


ETC.tsv
```diff
@@ -10375,7 +10375,7 @@ ETC_20150323_010376	Detect strange aura
 ETC_20150323_010377	{memo X}The place to bury the bottle
 ETC_20150323_010378	Metiken appearance trigger
 ETC_20150323_010379	{memo X}The place to install the magic field
-ETC_20150323_010380	Polluted pile of soil
+ETC_20150323_010380	Corrupted Pile of Soil
```
ttgmichael: You sure it's a corrupted pile of soil? there's somes quest lines I saw about piles of soil

vtange: iirc there are "corrupted/evil soil" in places like Gele Plateau (where you drag this doll thing around)

terrapin47: That one is currently referred to as "demon corrupted land"

vtange: aight. better safe than sorry

Sunaries:
:+1:

imcgames:
Thank you :smile:

PR #1239, opened 2015-10-15
Great Cathedral

terrapin47:


ETC.tsv
```diff
@@ -1698,7 +1698,7 @@ ETC_20150317_001697	Place the item you want to enhance.
 ETC_20150317_001698	{memo X}Card Enhancement
 ETC_20150317_001699	Gem Enhancement
 ETC_20150317_001700	You can no longer equip it. 
-ETC_20150317_001701	The item has been destroyed due to the death penalty.
+ETC_20150317_001701	Your equipment has been damaged due to the death penalty.
```
Sunaries, line 1701: Equipment -> item 

AFAIK, it's not always equipment. I've seen this message when losing my gems. That's an item and not equipment. Calling it item makes it neutral since equipment can be items.

terrapin47, line 1701: I don't think this message is notifying about your gems, but about your equipment only.  You drop your gems on death, but that's not what this message is saying.

Sunaries, line 1701: It refers to both: your gems being dropped and the durability of your items being lowered. 

I've heard that it also drops anvils, but that would require testing to see if it really does that.

terrapin47, line 1701: If it's both, then shouldn't it say "Your equipment has been damaged and your gems have dropped on the ground due to the death penalty."

Sunaries, line 1701: "Your items" is shorter. Damaged can mean that the gems fell on the ground

terrapin47, line 1701: Items implies that all your items were damaged, not just your equipped items.  I could say "equipped items".  I don't really see how "damaged" can mean "your gems fell on the ground".

Sunaries:
Good to merge.

imcgames:
:+1:

PR #1251, opened 2015-10-17
ETC

BloodHeartless:
little things


ETC.tsv
```diff
@@ -15128,7 +15128,7 @@ ETC_20151001_015129
 ETC_20151001_015130	
 ETC_20151001_015131	
 ETC_20151001_015132	
-ETC_20151001_015133	
+ETC_20151001_015133	You already found the required
```
Sunaries: Change it to: You already found what you need

ETC.tsv
```diff
@@ -10044,7 +10044,7 @@ ETC_20150323_010045	{memo X}had the item repaired.
 ETC_20150323_010046	$Collecting Young Leaves
 ETC_20150323_010047	$Collected a Young Leaf, but the tough leaf cut you!
 ETC_20150323_010048	$The tough leaves cut your hand!
-ETC_20150323_010049	You threw the music box back to where it was
+ETC_20150323_010049	You threw the music box to where it was
```
terrapin47, line 10047: This is something the player does after picking up the music box, they put it back.  It should stay as it was.

BloodHeartless, line 10047: pleonasm: back to where it was

terrapin47, line 10047: "to where it was" alone lacks something.  You don't say that in natural English.  You'd say "to where it was obtained" or "back to where it was"
If you want it shorter, you could say "You threw the music box back"

PR #769, opened 2015-08-09
Update UI

Sunaries:
Minor fix. Sorry for the many Pull Requests. It's just that I discover them afterwards. ^^;;

PR #1271, opened 2015-10-21
minor name fixing for #8265 & #8416

merrun:
8265 - change based on #176 in Item
8416 - change based on #769 in Item


ttgmichael:
:+1:

PR #1272, opened 2015-10-21
Minor edit for #8264 & #8416 as well as translate for #15673 

merrun:
따스한 오두막 to Cozy Lodge
8265 - change based on #176 in Item
8416 - change based on #769 in Item


Sunaries:
:+1:  (though Prieblanda fix is already in, but it doesn't matter)

imcgames:
:+1:

PR #1296, opened 2015-10-27
grammar fixes

Renari:
I'm not sure why the Korean text says it was changed, it was not.


Sunaries:
Miners' or Miner's. Either work, but we've already decided to stick with Miners' Village. Please remove the Korean text, it should remain blank unless it is translated in English. The rest is fine.

Renari:
Miners' is equivalent to Miners's which doesn't make sense here because the s in Miner's is already showing possession. The only time you would use miners is when you're referring to a group of people that have a profession that is mining. However when used in the title of a location like it is here it is implying that the village belongs to the miners thus Miner's is correct.

If you would still like me to revert the change however I will.

terrapin47:
Either is correct, but different.  If you're saying `Miner's` then you're saying that this is the place of everyone who has the profession of Miner.  Which would work, but would be a bit more appropriate for something like a Guild of learning, `Mage's Workshop` etc.

Saying `Miners'` refers to a specific group, this is the village of a group of miners.  This is slightly more appropriate here, since there's only one group of miners that belong here.  Again, either works in the end,  but there's no reason to change it.

Renari:
That's not exactly true, apostrophe s shows possession it has nothing to do with the amount of people. It just means that the village is their property. Miners' means the same exact thing, you use an apostrophe like that when something naturally ends in an s and you still want to show possession. For example if we had a person named Amias this is when you would use an apostrophe like shown in the title (That's Amias' sword). The word miner however does not naturally end in an s so if you are trying to show possession you would write miner's if you are not trying to show possession (meaning the village does not belong to the miners) then miners is correct. Grammatically miners' means you want to show possession but if that's the case you should have just put the apostrophe before the s to begin with instead of writing miners's which sounds very strange when pronounced due to a double consonant.

ttgmichael:
This isn't really a grammar fix PR. Miners' Village works fine already. Several examples of the ' being after the s (which means that the village belongs to a specific group) shown below:

http://finalfantasy.wikia.com/wiki/Cultists%27_Tower
http://finalfantasy.wikia.com/wiki/Pirates%27_Hideout
http://finalfantasy.wikia.com/wiki/Ancients%27_Maze
http://finalfantasy.wikia.com/wiki/Vikings%27_Cove
http://finalfantasy.wikia.com/wiki/Crawlers%27_Nest

And no, apostrophes come after the s for plural nouns as well (not just names that end with s like Amias) https://owl.english.purdue.edu/owl/resource/621/01/

three days' journey = journey of three days

Hence, the use of Miners' to show their possession is already grammatically correct.

stvkno:
I agree with Miners'. It is simple to google, and unless the usage is something like 'Miners Association', which means it is an association of more than one miner, it is always showing as "Miners'"

Miner's Village implies that it is one miner who owns the village, or one miner who is from the village. 

"We are going to the miner's village" - we are going to the village from whence the miner came
"We are going to the miners' village" - we are going to the village from whence the miners came.

Renari:
That's not exactly correct either apostrophe s can be used to refer to something that belongs to a group. Miners' isn't incorrect but it is very uncommonly used in English and a lot of people are going to look at it funny.

Sunaries:
Good to merge. :+1:

QUEST_LV_0100.tsv
```diff
@@ -355,13 +355,13 @@ QUEST_LV_0100_20150317_000350	$I don't know why they suddenly started to prey on
...
-QUEST_LV_0100_20150317_000356	$Awesome. You got rid of it by yourself.{nl}Go ahead and report to Aras. He'll be very pleased.
-QUEST_LV_0100_20150317_000357	$You mean you really defeated the Vubbe Fighter?{nl}Well then, that's a relief. Now the number of monsters will no longer increase.
+QUEST_LV_0100_20150317_000356	$Amazing, you got rid of it by yourself?{nl}Report to Aras, he'll be very pleased to hear this.
+QUEST_LV_0100_20150317_000357	$You really defeated the Vubbe Fighter alone!{nl}What a relief, now the number of monsters will stop increasing.
 QUEST_LV_0100_20150317_000358	$Of course. As promised, I will grant you access to the Miners' Village.{nl}Please wait.
 QUEST_LV_0100_20150317_000359	$If you find any clues, let the Operations Officer in the northern area know about it.
-QUEST_LV_0100_20150317_000360	$I think the battle at the Miners' Village is in a very bad situation.{nl}It's too late to move the troops so please go on ahead to support the Miners' Village.
+QUEST_LV_0100_20150317_000360	$The Miners' Village is in a dire situation.{nl}It's too late to organize the troops so please go ahead and support the Miners' Village.
```
vtange: still this instance of Miners' Village

ttgmichael, line 364: We've already come to agreement that Miners' Village is the name to use. Not Miner's Village.

vtange, line 364: woops, my bad, This is good then

Sunaries:
Good to merge. :+1:

imcgames:
:+1:

PR #1318, opened 2015-11-02
Verbage Updage

Ritiosk:
Reworded some item descriptions to make them less "Bulky" with less repetition between items.

### DIFF
--- a/ITEM.tsv
+++ b/ITEM.tsv
@@ -10,7 +10,7 @@ ITEM_20150317_000009	$Tasseled Armband
 ITEM_20150317_000010	$Empty_weapon 
 ITEM_20150317_000011	{memo X}$Crude Gladius 
 ITEM_20150317_000012	$One-Handed Sword [Slash] 
-ITEM_20150317_000013	$A weapon that is commonly used and carried by soldiers.
+ITEM_20150317_000013	$The standard weapon of the kingdom's soldiers that is also commonly carried by civillians.
 ITEM_20150317_000014	$Gladius 
 ITEM_20150317_000015	$This weapon is an improved version of its military counterpart.
 ITEM_20150317_000016	{memo X}$Low Falchion 
@@ -22,7 +22,7 @@ ITEM_20150317_000021	$Recommended for those who desire a heavier and stronger sw
 ITEM_20150317_000022	$Sabre 
 ITEM_20150317_000023	$Primary weapon of the Cavalry, optimized for slashing. The guard on the hilt helps protect the user's hand. 
 ITEM_20150317_000024	$Shamshir 
-ITEM_20150317_000025	$A blade that has a long, curved crescent. It dishes out heavy slashes.
+ITEM_20150317_000025	$The blade's large crescent edge cuts with immense power.
 ITEM_20150317_000026	$Scimitar 
 ITEM_20150317_000027	$Its heavy weight is compensated by its very powerful slashes. 
 ITEM_20150317_000028	$Kris 
--- a/ITEM.tsv
+++ b/ITEM.tsv
@@ -10,9 +10,9 @@ ITEM_20150317_000009	$Tasseled Armband
 ITEM_20150317_000010	$Empty_weapon 
 ITEM_20150317_000011	{memo X}$Crude Gladius 
 ITEM_20150317_000012	$One-Handed Sword [Slash] 
-ITEM_20150317_000013	$A weapon that is commonly used and carried by soldiers of the kingdom. 
+ITEM_20150317_000013	$A weapon that is commonly used and carried by soldiers.
 ITEM_20150317_000014	$Gladius 
-ITEM_20150317_000015	$This weapon is crafted with room for improvement over its military version. 
+ITEM_20150317_000015	$This weapon is an improved version of its military counterpart.
 ITEM_20150317_000016	{memo X}$Low Falchion 
 ITEM_20150317_000017	$This sword is exceptional for slashing. 
 ITEM_20150317_000018	{memo X}$Low Kaskara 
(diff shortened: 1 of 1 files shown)


ITEM.tsv
```diff
@@ -10,9 +10,9 @@ ITEM_20150317_000009	$Tasseled Armband
 ITEM_20150317_000010	$Empty_weapon 
 ITEM_20150317_000011	{memo X}$Crude Gladius 
 ITEM_20150317_000012	$One-Handed Sword [Slash] 
-ITEM_20150317_000013	$A weapon that is commonly used and carried by soldiers of the kingdom. 
+ITEM_20150317_000013	$A weapon that is commonly used and carried by soldiers.
```
ttgmichael: A weapon that is commonly used and carried by the kingdom's soldiers.

ITEM.tsv
```diff
@@ -10,9 +10,9 @@ ITEM_20150317_000009	$Tasseled Armband
...
 ITEM_20150317_000014	$Gladius 
-ITEM_20150317_000015	$This weapon is crafted with room for improvement over its military version. 
+ITEM_20150317_000015	$This weapon is an improved version of its military counterpart.
```
ttgmichael, line 15: This weapon is a marginally improved version of its military counterpart.

ITEM.tsv
```diff
@@ -22,11 +22,11 @@ ITEM_20150317_000021	$Recommended for those who desire a heavier and stronger sw
 ITEM_20150317_000022	$Sabre 
 ITEM_20150317_000023	$Primary weapon of the Cavalry, optimized for slashing. The guard on the hilt helps protect the user's hand. 
 ITEM_20150317_000024	$Shamshir 
-ITEM_20150317_000025	$The large curved crescent edge allows for a powerful slash. 
+ITEM_20150317_000025	$A blade that has a long, curved crescent. It dishes out heavy slashes.
```
ttgmichael: The old version is more accurate/closer to the Korean text.

Ritiosk: Looking over it, however, there seems to be multiple instances of repition between items. Just under this, on Scimitar we see the words "a powerful slash" in the exact same place.

And while it might be closer to the Korean text, I feel as though stylistically players would like to see wording that goes along moreso with the English that can be found written in Fantasy novels and the like :O

ITEM.tsv
Ritiosk: It could even be shortened to:
A weapon that is commonly used by the kingdom's soldiers. :astonished:

ttgmichael: Actually, I checked the full meaning of the Korean text is:

> The standard weapon of the kingdom's soldiers that is also commonly carried by civilians.

It's good to shorten descriptions, but we also need to keep the details that are in the Korean text ;).

ITEM.tsv
ttgmichael: Ok, so more variety, we could try to merge into one sentence though (since the korean text is 1 sentence):  

$The blade's greatly curved crescent edge allows it to dish out mighty slashes.

Ritiosk: And take out 'very very' words to strengthen the sentence's overall meaning

$The blade's crescent edge allows it to dish out might slashes. (?)

ttgmichael: I'm trying to incorporate as much of the korean text as possible. 

The Korean text literally is
"Has a large curved blade edge that can cut with immense power."

So would this work? (modifying from what you have above):

> $The blade's large crescent edge cuts with immense power.

Ritiosk: I'm in agreement with that one :+1:

ITEM.tsv
```diff
@@ -22,11 +22,11 @@ ITEM_20150317_000021	$Recommended for those who desire a heavier and stronger sw
...
 ITEM_20150317_000026	$Scimitar 
-ITEM_20150317_000027	$Its heavy weight is compensated by its very powerful slash. 
+ITEM_20150317_000027	$Its heavy weight is compensated by its very powerful slashes. 
 ITEM_20150317_000028	$Kris 
-ITEM_20150317_000029	$The twisted edge amplifies the damage it inflicts. A sword that is both practical and decorative. 
+ITEM_20150317_000029	$A snakelike sword that is both practical and decorative.
```
ttgmichael, line 29: As for this line, the original line is also closer with the Korean text.

We can stick with snakelike vs twisted, but it does say the edge of the sword magnifies/amplifies the damage it inflicts.

ttgmichael:
Could you reedit the lines so that they match the ones we discussed? Go to https://github.com/Ritiosk/EnglishTranslation/tree/patch-1 and you can edit the ITEM file from there. Else, I could make the changes on my PR.

ITEM.tsv
Ritiosk, line 29: Ah, I guess it's a personal preference thing for me when it comes to things that "Amplify" or have "Powerful" in the description text. If it actually comes with a bonus to those things... Such as something that increases damage to Beast mobs, it could say "Amplifies the damage it inflicts to beast-type monsters." but if it's just a basic weapon without the tertiary attributes, it shouldn't say that it's "Powerful" or that it "Amplifies" anything, etc.? 

Kind of like picking up a Lv. 1 sword that has "Powerful Slashes" and then a Lv. 150 sword that "Has a dull edge" etc. Not sure if you're following me :astonished:

Ritiosk:
Definitely re-editing the lines we discussed.

ITEM.tsv
ttgmichael, line 29: I get your point. You see the word amplify as more of a numerical verb. As in you take "damage" as a quantifiable measurement.

What the description is trying to say is that the winding edge inflicts a more significant wound/injury than a straight edge (so the sword makes a more serious looking injury, not necessarily anymore damaging if you quantify it).

Thus:

> Its snakelike edge deepens the wounds it creates, the sword is both practical and decorative.

Ritiosk, line 29: Aye, you have the right of it.

I'll work on updating this line in. 

Sorry I haven't worked on this more, school caught up with me :whale:

imcgames:
Updated with another PR

PR #1320, opened 2015-11-03
Talt "flavor" text

Sourpusss:
Added as a gaming joke for the notoriously annoying boss drop.  Feel free to ignore this request, but I've heard many players make this joke, so it'll probably be well received.

### DIFF
--- a/ITEM.tsv
+++ b/ITEM.tsv
@@ -2345,7 +2345,7 @@ ITEM_20150317_002344	Chinency Root
 ITEM_20150317_002345	{memo X}Infro Rocktor Bone 
 ITEM_20150317_002346	{memo X}It's not only the shell that's sturdy, but also the insides. Obtained from Infro Rocktors. 
 ITEM_20150317_002347	Talt 
-ITEM_20150317_002348	A strange gemstone sometimes found in monsters. Attracted by its mysterious color, many collectors seek this gem. 
+ITEM_20150317_002348	A strange gemstone sometimes found in monsters. Attracted by its mysterious color, many collectors seek this gem. One curious amasser reported the stone to have a very salty taste.
 ITEM_20150317_002349	Chupacabra Skin 
 ITEM_20150317_002350	Chupacabra Eyeball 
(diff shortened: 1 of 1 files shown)


vtange:
I don't get it, mind explaining?

Sourpusss:
Getting the currently useless talt makes people "salty" or "bitter" towards receiving the token, in comparison to equipment.

ttgmichael:
> Getting the currently useless talt makes people "salty" or "bitter" towards receiving the token, in comparison to equipment.

A strange gemstone sometimes found in monsters. Captivated by its mysterious color, many collectors seek this gem, while others wonder why they would choose to collect such a salty gemstone.

?

Sourpusss:
I like that way, but I don't really think the wording of "salty gemstone" works all that well.  It's not that the stone itself is feeling salty, and no one should really be tasting gemstones anyway, so it kind of explicitly needs to be said that someone tasted it (but it's not expected for anyone to do so).

Once again, no pressure on you guys, IMC.  This is more of a player joke than actual lore ideas.

ttgmichael:
Lol but the point of the meme is that out of all the flavors, you chose something salty xD. 
Just offering a suggestion :p

stvkno:
I don't think this is necessary at all.. Translations should be as close to the original as possible with grammar/context fixes and not personal notes/interpretations of the original text

Sourpusss:
You're right - it's not necessary.  It's just fun, and it gives the game some character.

Also, there's an arrow to the knee joke.  It's not like this is the first "extra" line added.

Sunaries:
Except the "arrow to the knee" was literally there. :wink:

Sourpusss:
Haha, well, kudos for IMC for the topical gaming humor.  Once again, this is just a frivolous suggestion based on current player feelings - just throwing ideas out there.

vtange:
if IMC wants to add the joke, then good to merge

imcgames:
I think it's good to merge. :+1:

PR #1345, opened 2015-11-09
Update SKILL.tsv

ArkKnight:


SKILL.tsv
```diff
@@ -5706,9 +5706,9 @@ SKILL_20151102_005705	$Increases the damage dealt on an enemy with [Incineration
 SKILL_20151102_005706	$Bloodletting: Hemostasis
 SKILL_20151102_005707	$Increases the cycle time of the HP cost by 0.2 seconds per attribute level while [Bloodletting] is active.
 SKILL_20151102_005708	$Bloodletting: Replace SP
-SKILL_20151102_005709	$Cuts SP instead of HP at a 10% chance per attribute level while the effect per tick of [Bloodletting] is active.
-SKILL_20151102_005710	$Increases the chance by 5% per attribute level to block Lv3 or less abnormal statuses for 10 seconds after receiving the effect of [Fumigate].
-SKILL_20151102_005711	$Purifies the area for 10 seconds within the range of [Fumigate]. Increases the Poison property resistance of allies by 10%.
+SKILL_20151102_005709	$While [Bloodletting] effect is active, each tick has a 10% chance per every attribute level to reduce SP instead of HP.
```
Sunaries, line 5709: While the [Bloodletting] effect is active, each tick has a 10% chance per attribute level to reduce SP instead of HP.

per / every - Both mean the same thing and so one of them is redundant. Use 'per' since that term is used elsewhere in SKILL.

SKILL.tsv
```diff
@@ -5706,9 +5706,9 @@ SKILL_20151102_005705	$Increases the damage dealt on an enemy with [Incineration
...
+SKILL_20151102_005710	$[Fumigate] effect grants you 5% chance per every attribute level to block level 3 or lower abnormal status for 10 seconds.
```
Sunaries: I don't see a problem with the previous line.

Sunaries, line 5710: What is wrong with the previous line? 

The KR string says it increases by 5% chance per attribute level after receiving the effect of Fumigate. Also, there are more abnormal statuses than just one.

PR #1348, opened 2015-11-10
Minor translation - 17251 - enhance gem error message, and 871- minor…

merrun:
… name fixing

17251 강화 할 수 없는 젬입니다. - This gem cannot be enhanced.

871 mentiwood > mantiwood
since the all the drop regarding mentiwood refers it to mantiwood instead of mentiwood

please don't conflict again 
if still conflict again later on ..I gave up ..

### DIFF
--- a/ETC.tsv
+++ b/ETC.tsv
@@ -868,7 +868,7 @@ ETC_20150317_000867	Red Luxodon
 ETC_20150317_000868	The reduction in size isn't the only thing that got better after this beast mutated.
 ETC_20150317_000869	Beeteros
 ETC_20150317_000870	It is simple and effective for Beeteros to camouflage. Please do not touch the swaying leaf as it can stab you.
-ETC_20150317_000871	Mentiwood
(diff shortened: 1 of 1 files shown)


Sunaries:
:+1: good to merge

ETC.tsv
```diff
@@ -17246,7 +17246,7 @@ ETC_20151102_017247	You have been restricted from chatting until {Time}.
 ETC_20151102_017248	You have been restricted from trading until {Time}.
 ETC_20151102_017249	The .ipf files are damaged and cannot be read. The client will now shutdown.
 ETC_20151102_017250	You cannot use the gem on this equipment.
-ETC_20151102_017251
+ETC_20151102_017251 This gem cannot be enhanced.
```
grillo-delmal: Careful with the spacing here, there should be a tab in between, not a space

merrun:
I did use tab ...but it become that

imcgames:
Thank you :+1:

PR #1353, opened 2015-11-11
fix some mistake

mamoruk:


Sunaries:
:+1: Nice catch. Thank you. Good to merge.

@imcgames Do you want 신기전 to be localized to 'Divine Arrow' or keep it as 'Singijeon'? Note that 'Gung Ho' is also in the game.

imcgames:
@Sunaries We want to use "Divine Arrow". Thank you :+1:

imcgames:
:+1:

PR #1355, opened 2015-11-12
Rozalija minor fix

ttgmichael:
I don't think her name should be spelled with two i's. 

If anyone has an objection, let me know xD

### DIFF
--- a/ETC.tsv
+++ b/ETC.tsv
@@ -9203,7 +9203,7 @@ ETC_20150317_009203	[Ranger Master]{nl}       Nemoken
 ETC_20150317_009204	[Swordsman Master]{nl}              Rashua
 ETC_20150317_009205	[Wizard Master]{nl}        Lucia
 ETC_20150317_009206	[Archer Master]{nl} Edmundas Tiller
-ETC_20150317_009207	[Cleric Master]{nl}       Rozaliia
+ETC_20150317_009207	[Cleric Master]{nl}       Rozalia
 ETC_20150317_009208	[Priest Master]{nl}       Boruble
 ETC_20150317_009209	[Krivis Master]{nl}       Herkus
 ETC_20150317_009210	{memo X}[Auctioneer]{nl} Demares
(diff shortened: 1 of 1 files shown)


Sunaries:
Doing a quick Google search shows me that the name Rozaliia really does exist.

ttgmichael:
Lol seems like its the Slavic spelling of Rosalia then? 

Alternatively, the Lithuanian version, if I'm correct is Rozalija? @Xenobius2501 
http://www.behindthename.com/name/rozalia

Thanks for checking! xD

Xenobius2501:
Yes, if you "adapt" a name ending with "ia" or "ie" so that it becomes "localized" in Lithuanian it will often end up with an extra "J". Like this one, for example: https://lt.wikipedia.org/wiki/Amelija_i%C5%A1_Monmartro

The "j" makes the resulting name sound like "Ameliya" (well, "Rosaliya" in our case). 
No double "i", tho. 

TL;DR: Rosalia is the basic name and Rosalija is the same name with a slight Lithuanian "accent".

ttgmichael:
@Xenobius2501 It appears the Korean pronunciation is Ro ja lli ya'; so we might be using a 'z' instead of an 's' for the name. Would that work? or should I just go with the 's'? xD

Xenobius2501:
Yes, using z instead of s would make it even closer to "Lithuanian" version.

ttgmichael:
@Sunaries are we ok to go by the Lithuanian version of spelling the name? If not, we can revert back to 'ii'.

Sunaries:
@ttgmichael Your call on this one, because I don't know about this one. Ask IMC?

ttgmichael:
@imcgames I guess you can decide whether to merge this.
This PR simply changes the way the Cleric Master's name is spelled. 

**Rosalia** is the English form of the name.
The current one in the game is **Rozaliia**, which is the Slavic (could be Russian, Ukrainian, or Czech) form of the name.
This PR changes the spelling to **Rozalija** which is the Lithuanian form of the name.

Personally, I like the spelling of Rozalija better than Rozaliia.
Thoughts?

imcgames:
@ttgmichael  It looks good better than Rozaliia. Good to merge. :+1:

PR #1358, opened 2015-11-13
Update UI.tsv

lifeiscontent:
I just did a search and replace for joypad to gamepad, and also did a tiny cleanup of simplifying the text in the beginning tutorial.

I feel like most people call "joypads", gamepads. I also noticed that there were "Joystick" words in there.

Is this intentionally different?


vtange:
I dunno, I'm more familiar with the word "joypad" than "gamepad".

I'd avoid editing joystick unless you're _sure_ it refers to joypads

lifeiscontent:
@vtange what about "controller"?

vtange:
That would actually be better, since a lot of people use 'controllers' to play the game. Common word, easier to Google if need be

lifeiscontent:
@vtange updated

vtange:
Nice. This should be good to merge then.

imcgames:
Good to merge :+1:

PR #1362, opened 2015-11-15
Clean #1332 + #1345 (ArkKnight) Edits - Squire Skill Renaming

vtange:
Can close #1332 and #1345 once this is merged.

IMC, I need to know where these UI lines appear in-game. My guess was that these were related to the Squire Class, but when I asked contributors who made Squire characters they said they couldn't find these lines. 

-UI_20150317_000598 {@st43}Repair Item{/}
-UI_20150317_000599 {@st41}List of items for Repair{/}
-UI_20150317_000600 {@st44}Item Repair{/}
-UI_20150317_000601 {@st42}Repair{/}
-UI_20150317_000602 {@st59}Repair as current settings{/}
-UI_20150317_000603 {@st59}Do not repair{/}

<hr>

Vote on what to call Squire Skill "Weapon Maintenance" and "Armor Maintenance" - skill that repairs AND buffs item for a period of time.

Candidates: Strengthen, augment, harden, reinforce, refurbish, revamp, restoration, mending, maintenance,


imcgames:
Well, as a matter of fact, I also failed to find these lines appear in game. But it's not issue for us and we want to use like current translated text.

We want to use easy words for global fans to understand our game even the original text has a hidden meaning. So Maintenance is better than the others. 

Ah, the more important thing is that all translated texts should be fit in the game when playing.

Thank you at all times. @vtange

vtange:
Cool. Thanks for that info. Might want to {memo X} those lines if need be.

I'll return those lines to maintenance as soon as possible

vtange:
Good To Merge.

imcgames:
:+1:

PR #1376, opened 2015-11-23
wrong memo x

mamoruk:
:dizzy_face:

### DIFF
--- a/ETC.tsv
+++ b/ETC.tsv
@@ -15346,7 +15346,6 @@ ETC_20151001_015347	You've restored the energies!
 ETC_20151001_015348	You've restored the energies!{nl}The energies form the crystal pole are coming into the body!
 ETC_20151001_015349	Capture a Hanaming!
 ETC_20151001_015350	Capture the Hanaming before the other players!
-ETC_20151001_015351	
 ETC_20151001_015352	The game has ended.
 ETC_20151001_015353	{@st42}Goal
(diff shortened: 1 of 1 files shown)


Sunaries:
Why are you removing all these memo x's? @imcgames added these since they are unused and do not appear in the game.

ttgmichael:
Yep, there's no need to remove the {memo X}'s they actually help us distinguish lines that are used from those that are unused. Makes translating ToS more efficient :)

mamoruk:
please check the ETC_kor.tsv file ....

ttgmichael:
@mamoruk IMC hasn't been adding memo X regularly on lines for the Korean file. I assure you that the lines that are memo X'd in the English files are properly memo X'd.

mamoruk:
= =i found ETC_20150317_008379  already inused but it sign memo X

ttgmichael:
Screenshot? 

Most lines though should be unused. It's best not to remove these tags as they are put there by IMC.

You can report this to @imcgames for them to update what needs to be memo'x :+1:

ttgmichael:
huh, interesting. 

@imcgames I'll let you figure this out ^^

mamoruk:
:dizzy_face: 
All other texts's 'memo X' I submited only exist in 'ETC.tsv'. However, 'ETC_20150317_008379' are correct in 'ETC_kro.tsv', so I'm not sure those texts been unused or not.

PR #1381, opened 2015-11-24
Mentally Crazy Fix

ttgmichael:
I'm posting this PR as a WIP until we finalize what we'd like to change the term into.

Currently it is my first suggestion, replacing "mentally sick" with "eccentric"

We can also go with: quirky, weird, crazy. Going with "loony" would have a subtle reference to other popular medias.

References to the Hippie movement have been suggested as another change, which I would go for "You might be mistaken as a flower child."

Let me know what people think! :+1:


ttgmichael:
So from Discord, we are ok with Flower Child?

Good to merge?

Sunaries:
+1

imcgames:
:+1:

PR #1390, opened 2015-11-28
Fortress of the Land, ... Standard

ttgmichael:
There has been numerous inconsistent naming of 대지의 요새. Upon settling with a name with everyone at Discord, this PR goes through most instances of 대지의 요새, to call it _Fortress of the Land_.

It also changes lines in Quest with the Unicode … to three periods for consistencies between files. 

There were changes to memo X lines because I felt it'd be good to change them properly as well to avoid future confusions.


Sunaries:
Good to merge. :+1:

imcgames:
Thank you :+1:

PR #1410, opened 2015-12-15
Star's -> Astral

vtange:
Star Tower -> Astral Tower, so we might want to change these to Astral to keep things in order.


Sunaries:
Good to merge.

imcgames:
:+1:

PR #1412, opened 2015-12-15
Reviewing Lines Part 2

ttgmichael:
Dolar signs on QUESTLV100 lines 700-1000 with some revisions.

Let me know your thoughts! :)


QUEST_LV_0100.tsv
```diff
@@ -1,3 +1,4 @@
+
```
Sunaries, line 1: Extra line? Needs to be removed.

Sunaries:
Dumbla Pond to Dumbal Pond - Was this also changed in everything else?

Sunaries:
What is the meaning of Dumbliai? The wiki does say 'dumbla' in the second column.

ttgmichael:
Dumbliai is the correct form of algae, I'm not sure if you can just shorten it to Dumbla https://lt.wikipedia.org/wiki/Dumbliai

https://translate.google.com/?ie=UTF-8&hl=en&client=tw-ob#en/lt/algae

Sunaries:
Good to merge.

imcgames:
:+1:

PR #1420, opened 2015-12-20
Sapper Skill Punji Stake Error

WatchGintama:
http://i.imgur.com/8sRKJTe.jpg

The skill says it takes 1 wood but it actually takes 3.  The Korean says so too. 
Changed Wood to Tree because that's what the item is named.

Also the skill does not knockback other players in PVP.  :(


Sunaries:
Good catch. Except "Tree" should be replaced with "Wood" since it doesn't make sense for a wooden log to be named a tree. In Korean it can mean both tree and wood.

Sunaries:
Good to merge. :+1:

imcgames:
:+1:

PR #1433, opened 2016-01-06
Quests and ETC proofreading

Sunaries:
Proofread & localized English text, typos etc.

The reason why there are so many additions and deletions is because the program I use for dialogue checking (Translation Helper) adds a tab space to all blank lines.


Sunaries:
Good to merge. :)

imcgames:
Thank you :+1:

PR #1455, opened 2016-02-01
Message removal notification

grillo-delmal:
When deleting a message, the delete dialog mentions the removal of "your group". 

The Korean text doesn't seem to design a target, and "your group" is out of the context of the dialog.

I also noticed that other lines near this one belong to the friend list grouping system, so I checked the "Remove Group" button to see if it uses the same string (considering that this might have been the root of the translation error), but that function doesn't have a confirmation dialog.

### DIFF
--- a/ETC.tsv
+++ b/ETC.tsv
@@ -2914,7 +2914,7 @@ ETC_20150317_002913	New Group
 ETC_20150317_002914	Block
 ETC_20150317_002915	Remove Group
 ETC_20150317_002916	Change Group Name
-ETC_20150317_002917	Are you sure you want to delete this message?
+ETC_20150317_002917	Are you sure you want to delete this?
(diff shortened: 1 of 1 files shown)


ETC.tsv
```diff
@@ -2914,7 +2914,7 @@ ETC_20150317_002913	New Group
 ETC_20150317_002914	Block
 ETC_20150317_002915	Remove Group
 ETC_20150317_002916	Change Group Name
-ETC_20150317_002917	Are you sure you want to delete your group?
+ETC_20150317_002917	Are you sure you want to delete this message?
```
Sunaries: "Are you sure you want to delete this?"

Sunaries:
Good to merge. :+1:

imcgames:
Long time no see @Grillo-del-mal, Thanks :+1:

PR #1450, opened 2016-01-25
Proofreading IMC's newest lines. Preparation for Steam TS.

Sunaries:
@imcgames 

The cleanup looks good. It was needed to remove all the {memo x}'s. 
Since you removed all the $ marks, I can only assume that this will no longer be supported anymore?

PR #1458, opened 2016-02-02
Steam Test Server - Update 3

Sunaries:
@imcgames This is starting to become frustrating. 

1) The old {memo x} lines that now exist in all files are used again? Because these lines are inaccurate, mistranslated and irrelevant. Hence it was {memo x} to begin with.

2) I only noticed this in SKILL.tsv, but I see reverts. Take a look at this: https://github.com/Treeofsavior/EnglishTranslation/pull/1450/files#diff-2836d424f5cb92d5d554af294cac4334 

↑ These lines (Demolition Rune etc.) were in the file, but now they are empty. Why? **This just makes it more time consuming for everybody.** I do not know if this also happens to other files, but if this is the case then I'm disappointed in how you manage it.
I've had to return the lines in this pull request, but I'm upset that they were removed while they are supposed to be **in**.

I will wait until this is sorted out.


Sunaries:
Good to merge.

Sunaries:
Good to merge.

imcgames:
Sorry about the revert issues. It will be sorted out soon. Thanks.

PR #1477, opened 2016-02-23
Minor typo

WatchGintama:
It's feels like... > It feels like


WatchGintama:
will be fixed in Spadow's PR

PR #1485, opened 2016-02-27
Update Lolopanther and Virtov set names

ghost:
Simply copied from item names, which were already translated.

I couldn't find the lines for the corresponding Earth Tower recipes though (like `제작서-롤로팬서 부츠`, `제작서-롤로팬서 보우`, etc)


Sunaries:
Good to merge. 

The recipes aren't in the current text files. Requires an update.

imcgames:
Thanks :+1:

PR #1499, opened 2016-03-05
Orsha Route check

grillo-delmal:
Some things we've found with @Anasu 

Good to go!, any feedback is appreciated :)

### DIFF
--- a/EnglishTranslation-master/ETC.tsv
+++ b/EnglishTranslation-master/ETC.tsv
@@ -13531,7 +13531,7 @@ ETC_20150717_013532	You do not have the required attribute to mount this compani
 ETC_20150717_013533	{ThemeName} ({Count} characters)
 ETC_20150717_013534	Like +1
 ETC_20150717_013535	{Name} likes {Who}!
-ETC_20150717_013536	xx {@st41b}Retrieval Item
+ETC_20150717_013536	{@st41b}Retrieved Items
 ETC_20150717_013537	Use
 ETC_20150717_013538	After smelling the food, an Infrogalas Mage rushed in
 ETC_20150717_013539	Lured Infrogalas Mage
--- a/EnglishTranslation-master/ETC.tsv
+++ b/EnglishTranslation-master/ETC.tsv
@@ -16914,7 +16914,7 @@ ETC_20151102_016915	Scout Magic Tower 3F
 ETC_20151102_016916	Scout Magic Tower 4F
 ETC_20151102_016917	Scout Magic Tower 5F
 ETC_20151102_016918	Move the Magic Concentration Orb onto the  Magic Circle of Verification
-ETC_20151102_016919	Rid the Smelly Cart
+ETC_20151102_016919	Burn down the Smelly Carts
 ETC_20151102_016920	Eliminate the Demons in Zone 2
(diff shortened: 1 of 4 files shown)


EnglishTranslation-master/ETC.tsv
```diff
@@ -13718,7 +13718,7 @@ ETC_20150729_013719	{nl}{nl}The lower part displays the total carry weight of yo
 ETC_20150729_013720	Available quests can be found in the Quest window.
 ETC_20150729_013721	Press {img F5 40 40} to view available quests {nl}or quests that are in progress.
 ETC_20150729_013722	{nl}{nl}You can restart abandoned quests by pressing the 
-ETC_20150729_013723	 button. Some quests cannot be abandoned.
+ETC_20150729_013723	 button appear on quests that cannot be abandoned.
```
sususuyo, line 13721: i don't think this change is necessary if this line follows up the line directly before it?

EnglishTranslation-master/QUEST_LV_0100.tsv
```diff
@@ -12237,8 +12237,8 @@ QUEST_LV_0100_20151224_012237	Bishop Urbonas is the pillar sustaining the people
 QUEST_LV_0100_20151224_012238	I've been asking around to try to figure out what happened but it seems as if the priests have all dissappeared as well.{nl}
 QUEST_LV_0100_20151224_012239	My men are all looking for the whereabouts of Bishop Urbonas.. {nl}But it's pretty much a dead-end at this point.{nl}
 QUEST_LV_0100_20151224_012240	If you're as good as the rumors make you out to be, you'll have no trouble in finding where Bishop Urbonas is.{nl}Just make sure nobody knows about this.. The lord doesn't want the people being any more agitated than they already are.
-QUEST_LV_0100_20151224_012241	Could you ask that person how much longer he is planning on making us stand here?
-QUEST_LV_0100_20151224_012242	It's been days since we've arrived here, and yet we still haven't set foot in Orsha.{nl}Why did they make us travel all this way just to make us stand around like this?
+QUEST_LV_0100_20151224_012241	Could you ask that person how much longer is he planning on making us stand here?
```
sususuyo, line 12240: i feel like this would be correct if the question was directed at the character:
`How much longer is he planning on making us stand here?`
or if the person speaking was making you speak specifically that question to a person unrelated:
`Could you ask that person "How much longer is he planning on making us stand here?"`
but the question is actually whether or not the character can ask "that person" or not

im sorry if this isn't really making sense, but to me this sounds off 😅

EnglishTranslation-master/ETC.tsv
grillo-delmal, line 13721: This line is not a follow up, is from the quest help dialog. This is how it actually looks:
image

we changed that to:
image

EnglishTranslation-master/QUEST_LV_0100.tsv
grillo-delmal, line 12240: Those lines are from people that are just complaining because they are on a line, and that person is actually besides you.

image

I don't think that it needs the quotes since it looks ok within the context

Sunaries:
Good to merge.

grillo-delmal:
Conficts resolved ;)

imcgames:
Thanks :+1:

PR #1520, opened 2016-03-14
Dark Surge -> Dark Theurge

ghost:
This line is used for both the warlock's skill and a monster name it seems.
Attributes are also translated as `Dark Theurge`.

Or should it really be `Surge` even though the skill is called `Warlock_DarkTheurge` internally?


Sunaries:
Good to merge. :+1: 

The Warlock is a class that is able to use magic to conjure evil spirits and controlling to aid in battle.
`Theurge` is correct since it is related to [theurgy](https://en.wiktionary.org/wiki/theurgy) which has the following definition: `A form of magic designed to allow for worship or conjuration of, or communication with spirits or deities.`

PR #1529, opened 2016-03-17
Warrant -> Orders (Question inside, though)

Sourpusss:
Warrant -> Orders

Also, I posted an issue, but hopefully someone will see this request here sooner:

How does one edit the larger files that are locked in their raw format?  For example ETC.tsv - I can't open it up and paste in like I usually do.  Another fix would be to just explain how to do a pull request for a single file?  Sorry, I'm quite rusty / not used to this aspect of github.


Sunaries:
Good to merge.

PR #1534, opened 2016-03-18
Updated 9571-9708 edit for ITEM.tsv

Sourpusss:
This is the revised version of my previous PR, using the newest ITEM.tsv and with the added {nl}'s from Jay's edit.

Still Lines 9571-9708

Closed the previous PR with edits over the previous ITEM.tsv


Sunaries:
Good to merge.

PR #1555, opened 2016-03-23
Fix

hiderdes:


Sunaries:
You are adding lines that don't exist in the current ETC_kor file. 
This cannot be merged yet because there is no update on this repository yet that reflects the source.

PR #1584, opened 2016-04-01
Update boss card names to match the names of monsters they summon.

ghost:


Sunaries:
Good to merge. Thank you.

PR #1585, opened 2016-04-01
Minor Fix

melodica97:
Monster name change


Sunaries:
Good to merge. 

I think I saw some of the changes before? It seems your branch is not entirely up-to-date, because `For Continuing Research` is already in.

PR #1587, opened 2016-04-01
Cleaning up some awkward text

Doddler:


EnglishTranslation-master/ETC.tsv
```diff
@@ -4265,7 +4265,7 @@ ETC_20150317_004264	Fool (haha)
 ETC_20150317_004265	The amount is low.
 ETC_20150317_004266	Moni
 ETC_20150317_004267	 key has been assigned to 
-ETC_20150317_004268	 to your quick slot.
+ETC_20150317_004268	 on your quick slot bar.
```
Sunaries, line 4268: on -> to

EnglishTranslation-master/ETC.tsv
```diff
@@ -19574,21 +19574,21 @@ ETC_20151224_019574	You've completed {DAYPLAYCOUNT} times out of {DAYMAXCOUNT} t
...
-ETC_20151224_019579	The monsters in Gogal District 2 has weakened!
-ETC_20151224_019580	The monsters in Gogal District 1 has weakened!
+ETC_20151224_019578	You have collected enough Mushrooms
+ETC_20151224_019579	The monsters in Exhaustion District 2 have weakened!
+ETC_20151224_019580	The monsters in Exhaustion District 1 have weakened!
 ETC_20151224_019581	The debuff effect wore off
 ETC_20151224_019582	The door to the left secret room has opened
 ETC_20151224_019583	The door to the right secret room has opened
-ETC_20151224_019584	Controlling the Device
+ETC_20151224_019584	The device has been activated
```
Sunaries, line 19583: Controlling the device (or Operating the device)

(The sentence in Korean implies you are performing an action; controlling the device to activate it)

EnglishTranslation-master/ETC.tsv
```diff
@@ -19574,21 +19574,21 @@ ETC_20151224_019574	You've completed {DAYPLAYCOUNT} times out of {DAYMAXCOUNT} t
...
 ETC_20151224_019585	The door of the main room has opened
 ETC_20151224_019586	The door of the main room is closed
-ETC_20151224_019587	This device is already being controlled.
+ETC_20151224_019587	This device has already been activated
```
Sunaries, line 19586: This device has already been operated

(do you actually activate it ingame?)

EnglishTranslation-master/ETC.tsv
Doddler, line 19583: So this line is from Nevellet Quarry, the 2f puzzle quest. Once you kill enough mobs near one of the two doors, the door opens, and there's a torch behind the door. This line is a map wide broadcast when you light one of the torches. Once you light both torches, the main door opens and you can receive a treasure reward. So in this context, 'controlling' doesn't make much sense. I understand my suggestion may not be the right answer, but the original is very unclear.

EnglishTranslation-master/ETC.tsv
Doddler, line 19586: This line appears as a map wide broadcast when someone attempts to light one of the torches that has already been lit.

EnglishTranslation-master/ETC.tsv
Doddler, line 4268: In game it would appear as "2 key has been assigned to Quick Cast on your quick slot bar" (for when it places quick cast in the slot assigned to the 2 key on your keyboard). Optimally it would make way more sense if the skill name and key could be swapped, but I think that's out of scope of the localization.

Sunaries, line 4268: You don't assign something _on_, but _to_.

EnglishTranslation-master/ETC.tsv
Sunaries, line 19583: _Activating the device_

Since the original implies an action being currently performed.

EnglishTranslation-master/ETC.tsv
Doddler, line 4268: Am I missing something? The text 'Assign to' is present and not changed. This change would read as "(keyname) key has been assigned to (skill name) on your quick slot bar."

Sunaries:
Good to merge.

PR #1621, opened 2016-04-23
Update some item and npc names for consistency

ghost:
Not sure about Windigo Searchers/Hunters though.
`Hunter` might sound better but it's only used in the Born with a Silver Spoon quest. `Searcher` is used everywhere else.


Sunaries:
Excellent. Good to merge. :+1:

imcgames:
Great :+1:

PR #1638, opened 2016-05-12
Earth Tower Prompt

FatePGN:
The prompt says "Choosing a party leader", when the party leader is
actually deciding on whether to stop at the current floor, or move on.

### DIFF
--- a/EnglishTranslation-master/ETC.tsv
+++ b/EnglishTranslation-master/ETC.tsv
@@ -20780,7 +20780,7 @@ ETC_20160224_020779	The mission for Earth Tower 10F will begin shortly
 ETC_20160224_020780	The first objective has begun{nl}Defeat Gosarius
 ETC_20160224_020781	Gosarius has called for his subordinates
 ETC_20160224_020782	Defeat Gosarius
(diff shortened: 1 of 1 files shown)


EnglishTranslation-master/ETC.tsv
```diff
@@ -20780,7 +20780,7 @@ ETC_20160224_020779	The mission for Earth Tower 10F will begin shortly
 ETC_20160224_020780	The first objective has begun{nl}Defeat Gosarius
 ETC_20160224_020781	Gosarius has called for his subordinates
 ETC_20160224_020782	Defeat Gosarius
-ETC_20160224_020783	Choosing a party leader
+ETC_20160224_020783	Party leader is deciding
```
Sunaries, line 20783: Can you change it to: "The party leader is deciding"

FatePGN, line 20783: Changed it

Sunaries:
Good to merge. :+1:

FatePGN:
Did another bit, where the refreshment table had incorrect translations for the food buffs, the HP recovery one was actually for SP, vice versa.

PR #1666, opened 2016-08-10
Change from madakia to medakia

calendulish:
The monster 메다키아 appears in translation as madakia and medakia
as well as your drops, but the correct translation for the term is
medakia.

Sample references: ETC_20150317_003290, ETC_20150317_007541

These commits fix all references (madakia -> medakia)


Sunaries:
Good to merge.

PR #1685, opened 2016-11-25
Small Fix

sunhwapark:
Some quest name adjustments, and few dialog fixes with some NPC names being centrally aligned once more


Sunaries:
Good to merge.

PR #1691, opened 2017-03-27
Skill Update

sunhwapark:
Skill typofix.
[Breaking Wheel] Skill description for Inquisitor seems to be missing, or i just couldn't find it.


sunhwapark:
`SKILL_20170313_007665` is what's being referenced by the client for the [`Frenzied Slash`] description, however the line below (`SKILL_20170313_007666`) is also a duplicate of this. For now, i've proposed the change of it to Frenzied slash, as i cannot find anything further about an ability named` Dragon Dive`.

sunhwapark:
Applied a change based upon a user's client data.

Attribute

PR #1697, opened 2017-11-26
English Translation Fixes

Crevox:


EnglishTranslation-master/ETC.tsv
```diff
@@ -26075,14 +26075,14 @@ ETC_20161214_026069	Reaction Speed Test Gimmick	순발력 테스트 기믹
 ETC_20161214_026070	{img green_up_arrow 16 16} 6 magic defense, {img red_down_arrow 16 16} 6 physical defense for 30 seconds with every basic and physical attack{nl}Increases Holy and Dark property attack equal to 7.5% magic defense for 10 seconds when stacked 20 times	기본 공격, [물리] 공격 시마다 8초간 포라포레 버프가 걸립니다.{nl}버프 100중첩 시, 10초간 자신의 마법방어력의 8%만큼 빛, 어둠 속성 공격력 증가	ETC_20151224_018669	{img green_up_arrow 16 16} 6 magic defense, {img red_down_arrow 16 16} 6 physical defense for 30 seconds with every basic and physical attack{nl}Increases Holy and Dark property attack equal to 7.5% magic defense for 10 seconds when stacked 20 times	기본 공격, [물리] 공격 시마다 30초간 마법 방어력 {img green_up_arrow 16 16}6 증가, 물리 방어력 {img red_down_arrow 16 16}6감소 효과{nl}효과 20중첩 시, 10초간 마법 방어력의 7.5%만큼 빛, 어둠 속성 공격력 증가	
 ETC_20161214_026071	Solmiki Cloth Set	솔미키 천 세트				
 ETC_20161214_026072	All Stats {img green_up_arrow 16 16}25	모든 스탯 {img green_up_arrow 16 16}25				
-ETC_20161214_026073	Max. SP {img green_up_arrow 16 16}1437 Increase	최대 SP {img green_up_arrow 16 16}1437 증가				
+ETC_20161214_026073	SP {img green_up_arrow 16 16}1437	최대 SP {img green_up_arrow 16 16}1437 증가
```
Sunaries, line 26078: I would avoid removing 'Max.' to avoid inconsistency. Other than that, this PR looks good. 👍

Crevox: Oh, I see what you mean about the KR text. Hmm. The problem is sometimes they leave it from the KR and sometimes they don't (like the word "Increase"). Items generally don't have "max." before them for the stats so I figured omitting it here made sense. I'll put it back for consistency with the other sets.

Crevox: After reviewing, the new accessory sets don't use "Max." or "Increase." This makes sense because sometimes it's there, sometimes it's not, but it never was on the original item properties outside of set bonuses. I've removed it from the two sets that still had it (Solmiki Cloth/Plate).

Crevox:
Should be good now. Please let me know if there's any other suggestions on changes.

Sunaries:
Good to merge. 👍 

Looking at your history, it seems you've submitted changes you've already done in the past. 

Ex)
- https://github.com/Treeofsavior/EnglishTranslation/commit/8f4ea22a3d7c932cfde9ea0f70e28467796a96cc
- https://github.com/Treeofsavior/EnglishTranslation/commit/18ab955c346757ae174a04f4338eefef0bf639d4
- https://github.com/Treeofsavior/EnglishTranslation/commit/2f54c485fabc3d40e37114f8de2f077897c3e44a

IMC has done this many times in the past, overwriting new strings with old strings. Too bad they haven't learned from that.

Crevox:
They got reverted by a failed steam merge. I already talked to IMC about this and they asked me to resubmit them.

https://github.com/Treeofsavior/EnglishTranslation/commit/576f9da895b4200aab81c26345ba6044db1a8fea

PR #1703, opened 2018-01-08
Minor Update

Sunaries:
Minor Update - Testing Github Desktop.


EnglishTranslation-master/ETC.tsv
```diff
@@ -25942,7 +25942,7 @@ ETC_20161214_025936	Soul Crystal (14 Days)	소울 크리스탈(14일)
 ETC_20161214_025937	Using this Soul Crystal will resurrect you in the same position you died. Upon pick-up, use within 14 days.	전투 불능 시 소울 크리스탈을 사용하여 제자리에서 부활할 수 있습니다. 획득 후부터 14일동안 사용 가능합니다.				
 ETC_20161214_025938	EXP Tome: 10 Minutes	경험의 서:10분				
 ETC_20161214_025939	Applies various effects to your character for 1 hour. Upon pick-up, use within 10 minutes. Right-click to use. Upon pick-up, use within 14 days.{nl} - Consumes time when you are connected to the game.{nl}[EXP Tome Effect]{nl} - +30% EXP from hunting monsters{nl} - Doubled STA recovery rate	마우스 우클릭 시 사용되며 캐릭터에 1시간 동안 효과가 적용됩니다. 획득 후부터 10분 동안 사용 가능합니다. 획득 후부터 14일동안 사용 가능합니다.{nl} - 캐릭터 접속 시에만 시간이 소진됩니다.{nl}[경험의 서 효과]{nl} - 몬스터 처치 시 획득 경험치 +30% 증가{nl}(PC방 혜택, 토큰 효과와 중첩 가능){nl}(경험치 카드 적용안됨){nl} - 스태미나 회복량 2배 증가				
-ETC_20161214_025940	Instance One Entry Voucher	인스턴스 던전 1회 입장권				
+ETC_20161214_025940	Instanced One Entry Voucher	인스턴스 던전 1회 입장권
```
Crevox: Why instanced? It's a +1 instance entry voucher. This one doesn't make much sense, and falls out of consistency with the other instance reset items.

EnglishTranslation-master/ETC.tsv
```diff
@@ -27608,14 +27608,14 @@ ETC_20170330_027602	Your companion is not in active mode.	컴패니언이 휴면
 ETC_20170330_027603	Cannot log out during the matching process.	매칭 요청 중에는 로그아웃 할 수 없습니다.				
 ETC_20170330_027604	Attribute point limit exceeded.	특성 포인트 보유 한도를 초과합니다.				
 ETC_20170330_027605	Not enough attribute points.	특성 포인트가 부족합니다.				
-ETC_20170330_027606	Cannot move to another map during the matching process.	매칭 요청 중에는 맵 이동을 할 수 없습니다.				
+ETC_20170330_027606	You cannot move to another map during a match request.	매칭 요청 중에는 맵 이동을 할 수 없습니다.
```
Crevox: This is not a match request, it's the matching process for TBL; the original translation was correct.

EnglishTranslation-master/ETC.tsv
```diff
@@ -27925,21 +27925,21 @@ ETC_20170605_027919	A package that contains a My Wedding Costume. Right-click to
 ETC_20170605_027920	Soul Crystal (4 Medals)	소울 크리스탈 (메달 4개)				
 ETC_20170605_027921	Using this Soul Crystal will resurrect you in the same position you died.{nl} Can be exchanged for 4 medals in the Exchange Shop.	전투 불능 시 소울 크리스탈을 사용하여 제자리에서 부활할 수 있습니다.{nl} 리사이클샵에서 메달 4개로 교환 가능합니다.				
 ETC_20170605_027922	Attribute Points 1,000 Ticket	특성 포인트 1,000				
-ETC_20170605_027923	Will increase attribute points by 1,000. Right-click to use. 	사용시 특성 포인트가 1,000 증가합니다. 마우스 오른쪽 클릭으로 사용합니다.				
+ETC_20170605_027923	Increases your Attribute Points by 1,000. Right-click to use. 	사용시 특성 포인트가 1,000 증가합니다. 마우스 오른쪽 클릭으로 사용합니다.				
 ETC_20170605_027924	Shining Stat Reset Potion 	빛나는 스탯 초기화 포션				
 ETC_20170605_027925	Shining Skill Reset Potion 	빛나는 스킬 초기화 포션				
 ETC_20170605_027926	Attribute Points Box (Rank S) 	특성 포인트 상자 S급				
-ETC_20170605_027927	A box that contains 10 Attribute Points 1,000 Tickets. Right-click to obtain. 	사용시 1,000 특성 포인트가 10장 들어있는 상자입니다. 마우스 오른쪽 버튼 클릭으로 사용하면 아이템이 지급됩니다.				
+ETC_20170605_027927	A box that contains ten Attribute Points 1,000 Tickets. Right-click to obtain. 	사용시 1,000 특성 포인트가 10장 들어있는 상자입니다. 마우스 오른쪽 버튼 클릭으로 사용하면 아이템이 지급됩니다.
```
Crevox: why change this? the original Korean text used the number.

EnglishTranslation-master/ETC.tsv
```diff
@@ -28777,7 +28777,7 @@ ETC_20170912_028771	/whale	/고래사냥
 ETC_20170912_028772	Opening this cube consumes {CONSUME}.{nl}Are you sure you want to open it?	큐브 개봉시{nl}{CONSUME}{nl}소모됩니다.{nl}정말 개봉하시겠습니까?				
 ETC_20170912_028773	Cannot be used near Goddess Statues.	여신상 주변에서는 사용하실 수 없습니다.				
 ETC_20170912_028774	Cannot be used near boss monsters.	보스 몬스터 주변에서는 사용하실 수 없습니다.				
-ETC_20170912_028775	Can only be used in fields.	필드에서만 사용하실 수 있습니다.				
+ETC_20170912_028775	You can only use this in fields.	필드에서만 사용하실 수 있습니다.
```
Crevox: This rule applies to everyone, not just you; this is a grammar change with no gain or loss, yet causes inconsistency with all of the other messages (like the two exactly above this) and increases the length of the line which causes risk.

EnglishTranslation-master/ETC.tsv
```diff
@@ -30161,7 +30161,7 @@ ETC_20171128_030155	Enter the Challenge Mode	챌린지 모드 입장
 ETC_20171128_030156	Step Up	스텝 업				
 ETC_20171128_030157	Finish the Mode	모드 종료				
 ETC_20171128_030158	Item Dismantle	아이템 분해				
-ETC_20171128_030159	Item Re-Identification	아이템 옵션 재감정				
+ETC_20171128_030159	Re-Identify Item	아이템 옵션 재감정
```
Crevox: this loses the consistency of "Item [action]" as seen in the line above

EnglishTranslation-master/QUEST.tsv
```diff
@@ -5449,7 +5449,7 @@ QUEST_20170303_005445	Enter monster level.{nl}(Max. value: 1 ~ 500)	몬스터를
...
 QUEST_20170313_005449	Demon Prison Guild (Recommended Lv : 110){nl}Only you and your party members can enter together.{nl}Are you sure you want to enter?	염원의 비석로(권장레벨:110)으로, 당신과 파티를 맺은 플레이어만이 함께 입장할 수 있습니다.{nl}입장하시겠습니까?	QUEST_20151224_003030	Demon Prison Guild (Recommended Lv : 115){nl}Only you and your party members can enter together.{nl}Are you sure you want to enter?	염원의 비석로(권장레벨:115)으로, 당신과 파티를 맺은 플레이어만이 함께 입장할 수 있습니다.{nl}입장하시겠습니까?	염원의 비석로
-QUEST_20170313_005450	Hello, I'm Identifier Sandra.{nl}Anything I can help you with?	안녕하세요. 저는 감정사 산드라입니다.{nl}제게 볼 일이 있으신가요?	QUEST_20161214_005002	Finding things that are hidden is always an interesting job.{nl}I'm sure there's something in it for you too.{nl}Alright, let me have a look.	숨겨진 걸 알아내는 건 언제나 즐거운 작업이에요.{nl}분명 당신께도 도움이 될 거에요.{nl}자! 제게 맡겨 보세요.	감정사 산드라
+QUEST_20170313_005450	Hello, I'm Sandra.{nl}Anything I can help you with?	안녕하세요. 저는 감정사 산드라입니다.{nl}제게 볼 일이 있으신가요?	QUEST_20161214_005002	Finding things that are hidden is always an interesting job.{nl}I'm sure there's something in it for you too.{nl}Alright, let me have a look.	숨겨진 걸 알아내는 건 언제나 즐거운 작업이에요.{nl}분명 당신께도 도움이 될 거에요.{nl}자! 제게 맡겨 보세요.	감정사 산드라
```
Crevox: the original translation included her title. There may be many Appraisers/Identifiers in the world, she's just specifying her title. This is not bad grammar, as people refer to themselves as "Doctor so-and-so" etc

EnglishTranslation-master/UI.tsv
```diff
@@ -2245,7 +2245,7 @@ UI_20160718_002244	{@st59}Shows dead parts of defeated monsters.{/}	{@st59}몬
 UI_20160718_002245	{@st66b}Show Character Silhouettes{/}	{@st66b}캐릭터 실루엣 적용{/}	UI_20160425_002180	{@st66b} Apply Character Silhouettes	{@st66b} 캐릭터 실루엣 적용	
 UI_20160718_002246	{@st59}Characters hidden behind the background are shown in silhouettes.{/}	{@st59}캐릭터가 배경 뒤에 있을 때 실루엣으로 표시합니다.{/}				
 UI_20160718_002247	Do you want to use this %s?{/}	%s 을 사용하시겠습니까?{/}	UI_20151224_001958	Do you want to use this %s?{/}	%s 아이템을 사용하시겠습니까?{/}	
-UI_20160718_002248	Unable to Use :	사용제한 : %s				
+UI_20160718_002248	Usable until :	사용제한 : %s
```
Crevox: until seems like a bad choice of wording, perhaps "Useable Time" or "Use Time"

EnglishTranslation-master/UI.tsv
```diff
@@ -2615,7 +2615,7 @@ UI_20170330_002614	{@st45tw2}Others{/}	{@st45tw2}기타{/}
 UI_20170330_002615	{@st42}Join	{@st42}참석하기				
 UI_20170330_002616	{@st42}Close	{@st42}닫기				
 UI_20170330_002617	{@st59}Close Message Box{/}	{@st59}메시지 함을 닫습니다{/}	UI_20150317_000017	{@st59}Close{/}	{@st59}경매장을 닫습니다{/}	
-UI_20170330_002618	{#62100c}Untradable items: %s	{#62100c}거래불가 개수 : %s개	UI_20150918_001804	Total Untradable: %s	총 거래 불가: %s개	
+UI_20170330_002618	{#62100c}Untradables: %s	{#62100c}거래불가 개수 : %s개	UI_20150918_001804	Total Untradable: %s	총 거래 불가: %s개
```
Crevox: This was fine as is, and always has been; untradeables is bad use of plural,  at best the I in items should be capitalized and that's it

EnglishTranslation-master/UI.tsv
```diff
@@ -2725,8 +2725,8 @@ UI_20170912_002724	{@st59}Growth{/}	{@st59}성장{/}
 UI_20170912_002725	{@st59}Exploration{/}	{@st59}탐험{/}				
 UI_20170912_002726	Team	팀명				
 UI_20170912_002727	Character	캐릭명				
-UI_20170912_002728	{img test_cooltime 24 24}Points reset on the 1st day of every month.{/}	{img test_cooltime 24 24} 매월 1일마다 점수가 초기화됩니다.{/}				
-UI_20170912_002729	{@st41b}Top Ranks	{@st41b}상위 랭킹				
+UI_20170912_002728	{img test_cooltime 24 24}Points reset on the first day of every month.{/}	{img test_cooltime 24 24} 매월 1일마다 점수가 초기화됩니다.{/}				
+UI_20170912_002729	{@st41b}Top Ranking	{@st41b}상위 랭킹
```
Crevox: this puts inconsistency with the line below

EnglishTranslation-master/UI.tsv
```diff
@@ -2814,16 +2814,16 @@ UI_20170912_002813	{@st66}News	{@st66}공지
 UI_20170912_002814	{@st66}Notes	{@st66}소개글				
 UI_20170912_002815	News	공지글				
 UI_20170912_002816	Notes	소개글				
-UI_20170912_002817	Member {img %s 13 13}	길드원{img %s 13 13}				
+UI_20170912_002817	Name {img %s 13 13}	길드원{img %s 13 13}
```
Crevox: this is not the translation, and furthermore, is fine as is; it's specifying who the member is

EnglishTranslation-master/UI.tsv
```diff
@@ -2814,16 +2814,16 @@ UI_20170912_002813	{@st66}News	{@st66}공지
 UI_20170912_002814	{@st66}Notes	{@st66}소개글				
 UI_20170912_002815	News	공지글				
 UI_20170912_002816	Notes	소개글				
-UI_20170912_002817	Member {img %s 13 13}	길드원{img %s 13 13}				
+UI_20170912_002817	Name {img %s 13 13}	길드원{img %s 13 13}				
 UI_20170912_002818	Level {img %s 13 13}	레벨{img %s 13 13}				
-UI_20170912_002819	Position {img %s 13 13}	직위{img %s 13 13}				
+UI_20170912_002819	Memo {img %s 13 13}	직위{img %s 13 13}
```
Crevox: this is not the translation, and position makes sense to allow the guild leader to specify the name of a "rank" or "position" within the guild hierarchy

EnglishTranslation-master/UI.tsv
```diff
@@ -2854,15 +2854,15 @@ UI_20170912_002853	 of your Market fees of the previous week.{nl}ⓘ Spot reward
 UI_20170912_002854	Tuesday 06:00	화요일 06:00				
 UI_20170912_002855	{@st43b}{s18}Guild Emblem{/}	{@st43b}{s18}길드 엠블럼{/}				
 UI_20170912_002856	{@st41b}{nl}Image %s{/}	{@st41b}{nl}이미지 %s{/}				
-UI_20170912_002857	ⓘ The first emblem can be registered for free.{nl}ⓘ Changing the emblem costs %s0,000 Silver.{nl}ⓘ Your image must be sized %s and in %s format.{nl}ⓘ The file name must be composed of only alphanumeric characters.{nl}ⓘ The use of problematic images can result in restrictions.	ⓘ 길드 엠블럼의 최초 등록은 무료입니다.{nl}ⓘ 길드 엠블럼 변경시 %s만 실버가 사용됩니다.{nl}ⓘ 길드 엠블럼은 %s 크기의 %s 파일만 등록 가능합니다.{nl}ⓘ 파일 이름이 영문과 숫자로 이루어진 파일만 등록 가능합니다.{nl}ⓘ 문제 있는 이미지 사용 시 제재가 있을 수 있습니다.				
+UI_20170912_002857	ⓘ The first emblem can be registered for free.{nl}ⓘ Changing the emblem costs %s0000 silver.{nl}ⓘ Your image must be sized %s and in %s format.{nl}ⓘ The file name must be composed of only alphanumeric characters.{nl}ⓘ The use of problematic images can result in restrictions.	ⓘ 길드 엠블럼의 최초 등록은 무료입니다.{nl}ⓘ 길드 엠블럼 변경시 %s만 실버가 사용됩니다.{nl}ⓘ 길드 엠블럼은 %s 크기의 %s 파일만 등록 가능합니다.{nl}ⓘ 파일 이름이 영문과 숫자로 이루어진 파일만 등록 가능합니다.{nl}ⓘ 문제 있는 이미지 사용 시 제재가 있을 수 있습니다.
```
Crevox: Are we just removing the comma here so it's harder to read?

EnglishTranslation-master/UI.tsv
```diff
@@ -2854,15 +2854,15 @@ UI_20170912_002853	 of your Market fees of the previous week.{nl}ⓘ Spot reward
...
+UI_20170912_002857	ⓘ The first emblem can be registered for free.{nl}ⓘ Changing the emblem costs %s0000 silver.{nl}ⓘ Your image must be sized %s and in %s format.{nl}ⓘ The file name must be composed of only alphanumeric characters.{nl}ⓘ The use of problematic images can result in restrictions.	ⓘ 길드 엠블럼의 최초 등록은 무료입니다.{nl}ⓘ 길드 엠블럼 변경시 %s만 실버가 사용됩니다.{nl}ⓘ 길드 엠블럼은 %s 크기의 %s 파일만 등록 가능합니다.{nl}ⓘ 파일 이름이 영문과 숫자로 이루어진 파일만 등록 가능합니다.{nl}ⓘ 문제 있는 이미지 사용 시 제재가 있을 수 있습니다.				
 UI_20170912_002858	{@st43b}{s18}Hangout Entry Settings{/}	{@st43b}{s18}아지트 입장 설정{/}				
 UI_20170912_002859	Allow Outsiders	외부인 입장 가능				
 UI_20170912_002860	{@st43b}{s18}Guild News{/}	{@st43b}{s18}길드 공지{/}				
-UI_20170912_002861	{@st66}{s18}Register{/}	{@st66}{s18}등록{/}				
+UI_20170912_002861	{@st66}{s18}Submit{/}	{@st66}{s18}등록{/}
```
Crevox, line 2861: Register is fine, you are registering your message on the board

EnglishTranslation-master/ETC.tsv
sunhwapark: `Instanced One Entry Voucher` -> `Instanced Dungeon One Entry Voucher`

EnglishTranslation-master/ETC.tsv
sunhwapark: Perhaps it could be misread as `A box that contains 10 attribute points` to the skimming reader, but even with the change it still has the same meaning, It should be alright.

EnglishTranslation-master/ETC.tsv
sunhwapark: How about, `This can only be used in fields.` ?

The same could be done to `ETC_20170912_028774` and `ETC_20170912_028773`
i.e `ETC_20170912_028774	This cannot be used near boss monsters.`

EnglishTranslation-master/ETC.tsv
sunhwapark: `a match request` - > `matching process`
I  think, `You cannot move to another map during matching process.` would be OK as an alternative.

EnglishTranslation-master/UI.tsv
sunhwapark: I would suggest, `Expires In:` 
based upon where it is used. (see below)

image

EnglishTranslation-master/ETC.tsv
Sunaries: The sentence could use a touch up due to the numerals. Like @sunhwapark mentioned, it could be misread. How about renaming the actual item to "Attribute Points (1,000) Ticket"? - Reads much better and less confusing instead of 1,000 tickets.

`A box that contains 10 Attribute Points (1,000) Tickets`

EnglishTranslation-master/ETC.tsv
Sunaries: Yes, it applies to everyone but the error message pops up when *you* interact with the item. Maybe it depends per item, I don't know.

EnglishTranslation-master/QUEST.tsv
Sunaries: Debatable. 

In my opinion, for a minor NPC, it is unnecessary to keep using the title as her title and name can already be seen above her dialogue. She's also the only person with this title as an NPC. There's also a line in ETC that says `Unidentified items can be identified at the Appraiser Sandra NPC`.

EnglishTranslation-master/UI.tsv
Sunaries: `Expires in` is actually better since the timer counts down.

EnglishTranslation-master/UI.tsv
Sunaries: 👍

EnglishTranslation-master/UI.tsv
Sunaries: Right, when people mentioned you could write anything in it I immediately thought of a short notice/memo. The Korean does mention 'Position' or 'Rank'. I changed it to 'Rank' since it is commonly used in hierarchical relationships in guilds. Let me know what you think. :)

EnglishTranslation-master/ETC.tsv
Crevox: I think appending the "You" to everything is excessive, as there are a ridiculous number of lines that would need to have that prefixed onto them, even ones you already edited below; why change one and not others? Why have it at all? It's an error message, and not the standard.

EnglishTranslation-master/UI.tsv
```diff
@@ -2816,7 +2816,7 @@ UI_20170912_002815	News	공지글
 UI_20170912_002816	Notes	소개글				
 UI_20170912_002817	Name {img %s 13 13}	길드원{img %s 13 13}
```
Crevox: This still needs to be reverted, it has been fine for months and is not the translation nor a desired localization

EnglishTranslation-master/QUEST.tsv
Crevox: She's not a minor NPC, she's the job master for Appraiser which is a very important NPC. If anything, the translation of "Identifier" in all cases should be verified if it should be "Appraiser."

EnglishTranslation-master/UI.tsv
Crevox: My concern with rank over position is it overuses the word "Rank" when describing something about someone. This could easily be confused for "class rank" when looking at a list of people, as in rank 8, rank 9, etc. Hence, position makes the most sense.

